PR #25, opened 2017-09-12
fix(email): fix delegated email responder

rfougier:


client/app/email-domain/email-domain.service.js
```diff
@@ -737,7 +737,7 @@ angular.module("services").service(
             return this.OvhHttp.get("/email/domain/delegatedAccount", {
                 rootPath: "apiv6",
                 params: {
-                    accountName,
+                    accountName: accountName || undefined,
```
jleveugle: Why do we need to send `undefined` to API ?

If `accountName` is not defined, we can just remove the parameter and don't it imo :)

rfougier: i changed the way to call API with facultatives params

client/app/email-domain/email-domain.service.js
```diff
@@ -730,14 +730,14 @@ angular.module("services").service(
 
         /**
          * Obtain Emails delegation
+         * @param {string|null} domain
          * @param {string|null} accountName
-         * @param {string} domain
          */
-        getDelegatedEmails (accountName, domain) {
+        getDelegatedEmails (domain = undefined, accountName = undefined) {
```
jleveugle: If you don't pass arguments, domain / accountName are already undefined, so no need to define them with undefined value, right ?

PR #31, opened 2017-09-13
fix(*): fix cloudDB display

rfougier:


client/app/resources/i18n/hosting/Messages_fr_FR.xml
```diff
@@ -78,6 +78,7 @@
    <translation id="hosting_dashboard_service_offer__90_PLAN" qtlid="196036">90plan</translation>
    <translation id="hosting_dashboard_service_offer_ASPFREE" qtlid="196049">aspfree</translation>
    <translation id="hosting_dashboard_service_offer_BUSINESS" qtlid="314110">business 2010</translation>
+   <translation id="hosting_dashboard_service_offer_CLOUDWEBX_1">cloud web</translation>
```
jleveugle: `Cloud web` ?

rfougier: So, what is the correct offer name in this case ?

jleveugle: `Cloud Web`

rfougier: So why all other offer names are in lower case ?

PR #56, opened 2017-09-19
Chore/refactored validator

FredericEspiau:
### Description of the Change

Needed to add Email validation and I took advantage of the opportunity to refactored the whole file

### DIFF
--- a/client/app/components/validator/validator.service.js
+++ b/client/app/components/validator/validator.service.js
@@ -1,128 +1,123 @@
-angular.module("services").service("Validator", [
-    function () {
-        "use strict";
+/* eslint-disable class-methods-use-this, max-len */
+{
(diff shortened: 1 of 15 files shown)


client/app/components/validator/validator.service.js
```diff
@@ -1,128 +1,123 @@
-angular.module("services").service("Validator", [
-    function () {
-        "use strict";
+/* eslint-disable class-methods-use-this, max-len */
+{
+    const IPV4_REGEX = /^(?:(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])\.){3}(?:25[0-5]|2[0-4][0-9]|1[0-9][0-9]|[1-9]?[0-9])$/;
```
antleblanc: Maybe `validator-js` dependency could help you?

Apollidore: Actually has to depart from `validator.js` as this service is to validate data according to OVH rules, not according to the specifications.

jleveugle: Why validate a IPv4 should be different in your case ? (excluding some IP for example ?)

If the API has custom rules to validate an IP. In this case, you should get the pattern to use directly by API, and don't manage it in the front-end.

What are the difference between an IP and an IP in this case ?

PR #66, opened 2017-09-22
fix(navbar):  navbar will display if no guide exist

FredericEspiau:
### Description of the Change

The menu crashed if the guides did not exist in the constants file.

### DIFF
--- a/client/app/components/manager-navbar/manager-navbar.module.js
+++ b/client/app/components/manager-navbar/manager-navbar.module.js
@@ -11,79 +11,86 @@ angular.module("App").run((LANGUAGES, managerNavbar, User, constants, translator
 
         managerNavbar.setExternalLinks(universes);
 
-        managerNavbar.setInternalLinks([
-            {
-                label: translator.tr("common_menu_support_assistance"),
-                subLinks: [
-                    {
-                        label: translator.tr("common_menu_support_all_guides"),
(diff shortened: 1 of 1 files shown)


client/app/components/manager-navbar/manager-navbar.module.js
```diff
@@ -11,79 +11,86 @@ angular.module("App").run((LANGUAGES, managerNavbar, User, constants, translator
...
+                } else {
+                    OtrsPopupService.toggle();
+                }
+            }
+        }, {
+            label: translator.tr("common_menu_support_list_ticket"),
+            url: "#/support"
+        }];
+
+        if (currentSubsidiaryURLs != null) {
```
AxelPeter: !== ?

Apollidore: `x != null` === `x != null && x != undefined`

jleveugle: I understand @AxelPeter feedback.

What do you think about this syntax : `if(!_.isEmpty(currentSubsidiaryURLs)) ` ? :)

FredericEspiau: I actually should have adopted this syntax to start with.

jleveugle: Can you fix these PR plz ? :)

PR #72, opened 2017-09-26
Fix/analytics tag

AxelPeter:
UXMWEB-3693


AxelPeter:
If we all have the same need in tracking, yes why not.
Or an extend for https://github.com/ovh-ux/ng-at-internet or https://github.com/ovh-ux/at-internet-ui-router-plugin

tremblaymath:
@AxelPeter it's the same tracking requirements for order, cloud, dedicated, web and telecom. We should extend  https://github.com/ovh-ux/ng-at-internet.

PR #83, opened 2017-10-02
fix(hosting): allow deleting ssl certificates

euhmeuh:
### Description of the Change

The legacy way of checking that you have a SSL certificate is through the route `/hosting/web/{serviceName}`, in the field "hasHostedSsl".

But it's not enough! If the hosting is more recent, this value will be false, and you need to check `/hosting/web/{serviceName}/ssl`.


client/app/hosting/general-informations/GENERAL_INFORMATIONS.controller.js
```diff
@@ -18,6 +18,8 @@ angular
         }
 
         canDeleteSSL () {
-            return this.$scope.hosting != null && this.$scope.hosting.hasHostedSsl && this.alreadyHasSSL();
+            // hasHostedSsl is true only for legacy SSL offers
+            // the new system is through the ssl API
+            return this.$scope.hosting != null && (this.$scope.hosting.hasHostedSsl || this.alreadyHasSSL());
```
jleveugle: What do you think about : 

`canDeleteSSL () => _.get(this, "$scope.hosting.hasHostedSsl") || this.alreadyHasSSL();`

euhmeuh: :+1:

PR #91, opened 2017-10-05
fix(translations): fix price ht label

AxelPeter:
UXMWEB-3919


AxelPeter:
@antleblanc Yes, it was only for DE (look at UXMWEB-3919)

PR #115, opened 2017-10-11
fix(otrs): fix otrs popup style

AxelPeter:
UXMWEB-3905

### DIFF
--- a/client/app/components/ovh-angular-otrs/ovh-angular-otrs.less
+++ b/client/app/components/ovh-angular-otrs/ovh-angular-otrs.less
@@ -1,11 +1,19 @@
 .otrs-popup {
+    height: 75% !important;
+    padding-top: 30px;
+    right: 30px;
+
     #headerPopup {
-        position: relative !important;
-        top: 0 !important;
+        position: absolute;
+        top: 0;
     }
 
-    #contentPopup {
-        height: auto !important;
+    #contentPopup { padding: 0 20px 20px 20px; }
+
+    .ovh-form-flat::after {
+        content: "";
+        display: block;
+        clear: both;
     }
 }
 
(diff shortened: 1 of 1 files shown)


client/app/components/ovh-angular-otrs/ovh-angular-otrs.less
```diff
@@ -1,11 +1,19 @@
...
     #headerPopup {
-        position: relative !important;
-        top: 0 !important;
+        position: absolute;
+        top: 0;
     }
 
-    #contentPopup {
-        height: auto !important;
+    #contentPopup { padding: 0 20px 20px 20px; }
```
antleblanc: shorthand:

```less
#contentPopup {
    padding: 0 20px 20px;
}
```

client/app/components/ovh-angular-otrs/ovh-angular-otrs.less
```diff
@@ -1,11 +1,19 @@
...
+
+    .ovh-form-flat::after {
```
antleblanc: Usage of class `.ovh-form-flat` isn't the best path to follow.
First you didn't have this dependency on Web, right?
Secondly this dependency is now deprecated.
@jleveugle Existing another option?

AxelPeter: Why ? It's the classname of the form inside `#contentPopup`. I could use `#otrsPopupForm`, but it's just to clear a float. Or put the clear on `#contentPopup`.

client/app/components/ovh-angular-otrs/ovh-angular-otrs.less
AxelPeter: I prefer to declare all, but it works too

jleveugle:
I will clos this PR because it's not the right place to do this fix.

We should update the otrs module and use the associate stylesheet :)

PR #120, opened 2017-10-17
set the right locale in the controller

euhmeuh:
Durations in the *Out of Memory* popup were not translated.

### DIFF
--- a/client/app/private-database/oom/private-database-oom.controller.js
+++ b/client/app/private-database/oom/private-database-oom.controller.js
@@ -11,6 +11,8 @@ angular.module("App").controller(
         }
 
         $onInit () {
+            moment.locale(this.$scope.selectedLanguage.value.split("_")[0]);
+
             this.productId = this.$stateParams.productId;
 
             this.NB_DAY_OOM = 7;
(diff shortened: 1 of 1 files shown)


client/app/private-database/oom/private-database-oom.controller.js
```diff
@@ -11,6 +11,8 @@ angular.module("App").controller(
         }
 
         $onInit () {
+            moment.locale(this.$scope.selectedLanguage.value.split("_")[0]);
```
FredericEspiau: Does the `$scope` always contain a `selectedLanguage` ? And does a locale always contain a `_` ? (I know that in Android, `fr` is a valid locale).

euhmeuh: I guess so, because that's how we're doin' in every other controller that needs *moment.js*. I'm pretty sure a more elegant solution would require some refactoring though.

FredericEspiau: A `languageService` would be nice I guess _whistles_

AxelPeter: In general, there is a refactoring to do with all the inherited scopes. 
But this kind of config doesn't belong in a controller, as it can be applicable everywhere in the app.

euhmeuh:
Fixed by  #130

PR #122, opened 2017-10-18
Remove summit banner

euhmeuh:
Thanks for this summit! :+1:


client/app/configuration/configuration.html
```diff
@@ -1,6 +1,6 @@
 <div class="container-fluid" id="main-container" data-ng-controller="UserCtrl">
     <div class="row">
-        <div class="col-xs-12 col-md-10">
+        <div class="col-xs-12 col-md-12">
```
antleblanc: -`col-md-12`

AxelPeter: well seen :)

jleveugle: -- (row + col) ;)

PR #129, opened 2017-10-27
Standardization

rfougier:


client/app/css/less/bootstrap/_popover.less
```diff
@@ -4,11 +4,11 @@
     z-index: 1040;
 
     .popover-menu {
-        & > button {
+        button {
```
AxelPeter: To add in https://github.com/ovh-ux/ovh-ui-kit-bs ?

client/app/css/source.less
```diff
@@ -71,4 +71,16 @@
     left: inherit;
     right: inherit;
     z-index: inherit;
+
+    .dropdown-menu li {
```
AxelPeter: To add in https://github.com/ovh-ux/ovh-ui-kit-bs ?

PR #130, opened 2017-10-30
Feat/centralized moment locale

FredericEspiau:
### Description of the Change

The locale setting of moment.js is now centralized. No need to have it set whenever needed.

It also works in case of language change.


client/app/components/user/user-lang.controller.js
```diff
@@ -1,22 +0,0 @@
...
-        $scope.selectedLanguage = translator.getSelectedAvailableLanguage();
-        $scope.availableLanguages = translator.getAvailableLanguages();
-
-        if (moment && moment.locale && $scope.selectedLanguage.value) {
-            moment.locale($scope.selectedLanguage.value.replace(/_/, "-"));
-        }
-
-        $scope.setSelectedLanguage = (newLanguage) => {
-            $scope.selectedLanguage = newLanguage;
-            localStorage.setItem("univers-selected-language", newLanguage.value);
```
jleveugle: Why do we remove the localStorage call ?

FredericEspiau: The whole file is deleted, it wasn't used at all

FredericEspiau: The whole file is deleted, it wasn't used at all

jleveugle: It's not my question :/

In others managers, we use univers-selected-language to determine the language at first loading.

https://github.com/ovh-ux/ovh-manager-telecom/blob/8dde0a6043aca2d83da1b85e726f29d93273b15a/client/app/app.js#L116

Why do we delete it in web manager ? :(

FredericEspiau: This file wasn't used at all, I deleted it because it had no use.

The `univers-selected-language` processing is done by `app.js`:

    let defaultLanguage = "fr_FR";

    if (localStorage["univers-selected-language"]) {
        defaultLanguage = localStorage["univers-selected-language"];
    } else {
        localStorage["univers-selected-language"] = defaultLanguage;
    }

FredericEspiau: And the language switching is done by the navbar:

    _(LANGUAGES)
                .filter((language) => _(language).has("name", "value"))
                .map((language) => ({
                    label: language.name,
                    click: () => {
                        localStorage["univers-selected-language"] = language.value;
                        window.location.reload();
                    },
                    lang: _(language.value).chain()
                        .words()
                        .head()
                        .value()
                }))

PR #131, opened 2017-10-30
Fix/domain transfert progress bar

FredericEspiau:
### Description of the Change

The progress bar object is not available right after operation creation, which resulted in an error message in the manager. The error message won't display anymore.

### DIFF
--- a/client/app/domain-operation/progress/domain-operation-progress.controller.js
+++ b/client/app/domain-operation/progress/domain-operation-progress.controller.js
@@ -1,60 +1,76 @@
-angular.module("App").controller("DomainOperationProgressCtrl", class DomainOperationProgressCtrl {
+angular
+    .module("App")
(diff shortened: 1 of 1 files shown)


client/app/domain-operation/progress/domain-operation-progress.controller.js
```diff
@@ -1,60 +1,76 @@
...
+                    this.domain = operation.domain;
+                    this.creationDate = operation.creationDate;
+                    this.doneDate = operation.doneDate;
+
+                    const fullProgressBar = { progress: 100 };
+
+                    return operation.status === "done" ? fullProgressBar : this.Operation.getProgressBar(operation.id);
+                })
+                .catch((error) => {
+                    if (!_(error).startsWith("The requested object")) {
```
jleveugle: We shouldn't do that :(

Why do you need this ?

PR #133, opened 2017-10-31
chore(assistance): update OTRS module to 3.0.1

euhmeuh:


AxelPeter:
Is the style in this file still necessary with the update: https://github.com/ovh-ux/ovh-manager-web/blob/chore/update-otrs/client/app/components/ovh-angular-otrs/ovh-angular-otrs.less ?

euhmeuh:
@AxelPeter Looks like the less file was an old style fix. I removed it and nothing changes.

PR #136, opened 2017-10-31
Fix/senegal menu not working

FredericEspiau:
## Menu now displays for Senegal
The top navbar wouldn't display in case where no guides were given to the subsidiary. Now the navbar is much more sturdy.

### DIFF
--- a/client/app/components/manager-navbar/manager-navbar.module.js
+++ b/client/app/components/manager-navbar/manager-navbar.module.js
@@ -10,7 +10,7 @@ angular
         function setMenuItemssToOtherUniverses (currentUser) {
             const universeMenuItems = Object.keys(constants.MANAGER_URLS)
                 .map((universeName) => ({
-                    icon: universeName === "labs" ? "fa fa-flask fs24" : null,
+                    icon: universeName === "labs" ? "fa fa-flask" : null,
                     label: universeName === "labs" ? null : translator.tr(`universe_univers-${universeName}_name`),
                     title: translator.tr(`universe_univers-${universeName}_name`),
                     universe: universeName,


client/app/components/manager-navbar/manager-navbar.module.js
```diff
@@ -1,70 +1,54 @@
...
+        return User.getUser()
+            .then((currentUser) => {
+                setMenuItemssToOtherUniverses(currentUser);
+                setGeneralMenuItems(currentUser);
+            });
+
+        function setMenuItemssToOtherUniverses (currentUser) {
+            const universeMenuItems = Object.keys(constants.MANAGER_URLS)
+                .map((universeName) => ({
+                    icon: universeName === "labs" ? "fa fa-flask fs24" : null,
```
antleblanc: `fs24` can be removed?

client/app/components/manager-navbar/manager-navbar.module.js
```diff
@@ -85,7 +69,74 @@ angular.module("App").run((LANGUAGES, managerNavbar, User, constants, translator
...
+        function getLanguageChangingMenuObject () {
+            const languageChangingMenuObject = _(LANGUAGES)
+                .filter((language) => _(language).has("name", "value"))
+                .map((language) => ({
+                    label: language.name,
+                    click: () => {
+                        localStorage["univers-selected-language"] = language.value;
+                        window.location.reload();
+                    },
+                    lang: _(language.value).chain()
```
jleveugle: `_(language.value).chain()` === `_(language.value)`

FredericEspiau: You should read the remaining of the code:
_(language.value).chain()
    .words()
    .head()
    .value()

If I didn't use `.chain()` I would have had to do:

`_.head(_.words(language.value))`

jleveugle: 3 vs 4 my bad :(

PR #152, opened 2017-11-03
feat(privateDb): handle db capabilities

euhmeuh:
### Description of the Change

Only show CRUD buttons of the database page for features that are available in the backend.
To know what features are available, we check the "capabilities" object returned by the API.

### Benefits

We can change the backend and the interface adapts to what is available.

### Possible Drawbacks

Some users might be surprised or confused not to see the expected buttons when using a database that does not support some actions. We may want to tell them why.


client/app/private-database/private-database.service.js
```diff
@@ -117,8 +117,12 @@ angular.module("services").service(
                         .then((_response) => {
                             database.serviceInfos = _response.data;
                             return database;
-                        })
-                        .catch((err) => this.$q.reject(err.data));
+                        });
+                })
+                .then((db) => {
+                    // reformat capabilities array as an object for easy access
+                    db.capabilities = _.mapKeys(db.capabilities, (c) => c.object);
```
FredericEspiau: `c` could have been named `capability` to enhance readability

client/app/private-database/private-database.service.js
```diff
@@ -117,8 +117,12 @@ angular.module("services").service(
                         .then((_response) => {
                             database.serviceInfos = _response.data;
                             return database;
-                        })
-                        .catch((err) => this.$q.reject(err.data));
+                        });
+                })
+                .then((db) => {
```
FredericEspiau: `db` could have been named `database` but I'm nitpicking

client/app/private-database/private-database.service.js
```diff
@@ -117,8 +117,12 @@ angular.module("services").service(
...
+                    // reformat capabilities array as an object for easy access
```
FredericEspiau: As we had a talk about comments yesterday, I feel like this is the good time to say that this kind of comment feels unnecessary

euhmeuh: I felt like I wanted to explain why I was adding a formatting step to the query, but I guess you're right, it's pretty obvious.

PR #171, opened 2017-11-09
Fix(hosting): add missing expected return value

rfougier:


client/app/hosting/module/add/hosting-module-add.controller.js
```diff
@@ -239,23 +239,25 @@ angular.module("App").controller("HostingModuleCreateCtrl", ($scope, $q, $stateP
...
-                        $scope.rootPathPrefix = "";
-                    }
-                    $scope.defaultInstallationPath = [$scope.rootPathPrefix, $scope.pathPrefix, $scope.model.path].join("");
-                })
-                .catch((err) => {
-                    Alerter.alertFromSWS($scope.tr("hosting_tab_DATABASES_configuration_create_step1_loading_error"), _.get(err, "data", err), $scope.alerts.main);
-                    $scope.loading.domains = false;
-                    $scope.resetAction();
-                });
+            return null;
```
jleveugle: A little thing, if we return a promise in a case, we should return a promise in all case no ?

Because if you do yourMethod().then(() => {}) ... You will have `.then is not a function` :/

What do you think about this ?

rfougier: There is no method in this case => inside $watch. So it cannot be called by an other method.
I added a return statement because of an eslint alert

jleveugle: Ok nice ;)

My question was more generic :)

FredericEspiau: Can't you get rid of the `$watch` then ?

PR #180, opened 2017-11-13
style(hosting): handle and translate cryptic error message from the backend

euhmeuh:
An error message from the backend is shown when a user tries to order an SSL certificate while the hosting is not configured correctly. This message is cryptic and not translated.

This change allows a translation of that message (while keeping the original one in parentheses) and helps the user understand what he/she needs to do.

This is obviously a rather dirty and temporary fix. We need a better error message from the backend.

### DIFF
--- a/client/app/hosting/ssl/order/hosting-order-ssl.controller.js
+++ b/client/app/hosting/ssl/order/hosting-order-ssl.controller.js
@@ -63,7 +63,7 @@ angular.module("App").controller(
                         return rtn;
                     })
                     .catch((err) => {
-                        this.Alerter.alertFromSWS(this.$scope.tr("hosting_dashboard_ssl_order_error"), err, this.$scope.alerts.main);
+                        this._handleSslError(err);
                         this.$scope.resetAction();
                     })
                     .finally(() => {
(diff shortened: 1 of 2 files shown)


client/app/hosting/ssl/order/hosting-order-ssl.controller.js
```diff
@@ -88,14 +88,22 @@ angular.module("App").controller(
...
                     .finally(() => {
                         this.$scope.loadSsl();
                         this.$scope.resetAction();
                     });
             }
         }
 
+        _handleSslError (err) {
+            let message = this.$scope.tr("hosting_dashboard_ssl_order_error");
+            if (err.message === "No attached domain with ssl enabled or no attached domain that redirect on hosting IPs, please use hosting IPs in your domain zone") {
```
jleveugle: Forget :p

We should display a generic message in french and the SSL error in english in parenthesis.

To translate error, we are talking about error code etc ! We should find a generic solution ;)

FredericEspiau: :+1:

client/app/hosting/ssl/order/hosting-order-ssl.controller.js
```diff
@@ -88,14 +88,22 @@ angular.module("App").controller(
 
                 this.Hosting.createSsl(this.$stateParams.productId, data.certificate, data.key, data.chain)
```
FredericEspiau: `return`

client/app/hosting/ssl/order/hosting-order-ssl.controller.js
```diff
@@ -88,14 +88,22 @@ angular.module("App").controller(
...
-                    .catch((err) => this.Alerter.alertFromSWS(this.$scope.tr("hosting_dashboard_ssl_order_error"), err, this.$scope.alerts.main))
+                    .catch((err) => this._handleSslError(err))
                     .finally(() => {
                         this.$scope.loadSsl();
                         this.$scope.resetAction();
                     });
             }
         }
 
+        _handleSslError (err) {
```
FredericEspiau: what do you think about this convention @jleveugle ?

PR #187, opened 2017-11-15
Feat/ Suggest to add an authorised IP when creating a DBaaS

euhmeuh:
It's not always obvious for the user that when creating a database (cloudDB) for the first time, the server won't be able to access it until an IP has been whitelisted.

With this change, we detect when there are no whitelisted IPs yet, and directly ask for one in the database creation form.

The user can always uncheck this suggestion if he/she doesn't want to be bothered with this for now, and do it manually later on.

We don't suggest this anymore when at least one IP has been configured.

### DIFF
--- a/client/app/private-database/database/add/private-database-database-add.controller.js
+++ b/client/app/private-database/database/add/private-database-database-add.controller.js
@@ -1,29 +1,29 @@
 angular.module("App").controller(
     "PrivateDatabaseAddBddCtrl",
     class PrivateDatabaseAddBddCtrl {
-        constructor ($scope, $q, $stateParams, Alerter, PrivateDatabase, Hosting) {
+        constructor ($scope, $q, $stateParams, Alerter, PrivateDatabase, Hosting, WhitelistService, Validator) {
             this.$scope = $scope;
             this.$q = $q;
             this.$stateParams = $stateParams;
             this.Alerter = Alerter;
             this.PrivateDatabase = PrivateDatabase;
             this.Hosting = Hosting;
-
-            this.bdd = {
-                value: null,
-                addUser: false,
-                addUserLogin: "",
-                addUserGrant: "",
-                addUserPassword: "",
-                addUserPasswordConfirm: ""
-            };
+            this.whitelistService = WhitelistService;
+            this.validator = Validator;
         }
 
         $onInit () {
             this.productId = this.$stateParams.productId;
 
+            this.checkAuthorizedIp()
+                .then((hasAuthorizedIp) => {
+                    this.model.addIp = !hasAuthorizedIp;
+                    this.hasAuthorizedIp = hasAuthorizedIp;
+                });
+
             this.model = {
                 addUser: false,
+                addIp: false,
                 database: {
                     value: "",
                     condition: {
@@ -58,6 +58,9 @@ angular.module("App").controller(
                             }
                         }
                     }
+                },
+                ip: {
+                    value: ""
                 }
             };
 
@@ -66,8 +69,13 @@ angular.module("App").controller(
             });
 
             this.$scope.addBdd = () => this.addBdd();
-
             this.$scope.isBddValid = () => this.isBddValid();
+        }
+
+        checkAuthorizedIp () {
+            return this.whitelistService
+                .getWhitelistIds(this.productId)
+                .then((result) => !_.isEmpty(result));
         }
 
         addBdd () {
@@ -85,6 +93,16 @@ angular.module("App").controller(
                         });
                     }
                     return this.PrivateDatabase.addBdd(this.productId, this.model.database.value);
+                })
+                .then(() => {
+                    if (this.model.addIp) {
+                        return this.whitelistService.createWhitelist(this.productId, {
+                            ip: this.model.ip.value,
+                            name: this.$scope.tr("privateDatabase_add_authorized_ip_description"),
+                            service: true,
+                            sftp: false
+                        });
+                    }
                 })
                 .then(() => this.Alerter.success(this.$scope.tr("privateDatabase_add_bdd_success"), this.$scope.alerts.main),
                       () => this.Alerter.error(this.$scope.tr("privateDatabase_add_bdd_fail"), this.$scope.alerts.main));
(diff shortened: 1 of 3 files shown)


client/app/private-database/database/add/private-database-database-add.controller.js
```diff
@@ -1,29 +1,31 @@
...
+            this.validator = Validator;
         }
 
         $onInit () {
             this.productId = this.$stateParams.productId;
 
+            if (this.$scope.isDBaaS()) {
+                this.checkAuthorizedIp()
+                    .then((hasAuthorizedIp) => {
+                        this.model.addIp = !hasAuthorizedIp;
```
FredericEspiau: Maybe you could rename `addIp` to something more explicit like `userCanAddIPAddress`

client/app/private-database/database/add/private-database-database-add.controller.js
```diff
@@ -86,6 +96,16 @@ angular.module("App").controller(
...
+                    if (this.model.addIp) {
+                        return this.whitelistService.createWhitelist(this.productId, {
+                            ip: this.model.ip.value,
+                            name: this.$scope.tr("privateDatabase_add_authorized_ip_description"),
+                            service: true,
+                            sftp: false
+                        });
+                    }
+                })
                 .then(() => this.Alerter.success(this.$scope.tr("privateDatabase_add_bdd_success"), this.$scope.alerts.main),
```
FredericEspiau: Extract the second method to put it into a `catch`

client/app/private-database/database/add/private-database-database-add.controller.js
```diff
@@ -120,5 +140,9 @@ angular.module("App").controller(
         getGrantLabel (grant) {
             return grant ? this.$scope.tr(`privateDatabase_add_bdd_new_user_grant_${grant}`) : "";
         }
+
+        isIpValid (ip) {
+            return this.validator.isValidIpv4Block(ip) || this.validator.isValidIpv4(ip);
```
FredericEspiau: Isn't `this.validator.isValidIpv4(ip)` enough ? If I remember correctly, it accepts CIDRs

client/app/private-database/database/add/private-database-database-add.html
```diff
@@ -119,6 +119,41 @@
                     </div>
                 </div>
             </form>
+
+            <form class="form-horizontal" name="privateDatabaseAddAuthorizedIpForm" ng-if="isDBaaS() && !addBddCtrl.hasAuthorizedIp">
```
FredericEspiau: This function isn't linked to `addBddCtrl` ?

client/app/private-database/database/add/private-database-database-add.html
```diff
@@ -119,6 +119,41 @@
...
+                            <span class="oui-icon oui-icon-checkbox-checked" aria-hidden="true"></span>
+                            <span class="oui-icon oui-icon-checkbox-checkmark" aria-hidden="true"></span>
+                        </span>
+                    </label>
+                </div>
+
+                <div class="alert alert-warning help-block mt-3" data-ng-if="addBddCtrl.model.addIp" data-i18n-static="privateDatabase_add_authorized_ip_explanation"></div>
+
+                <div class="mt-5" data-ng-if="addBddCtrl.model.addIp">
+                    <div class="form-group" data-ng-class="{'has-error': privateDatabaseAddAuthorizedIpForm.bddAddIp.$dirty && !addBddCtrl.isIpValid(addBddCtrl.model.ip.value)}">
```
FredericEspiau: You're mixing ui-kit and bootstrap in this view

client/app/private-database/database/add/private-database-database-add.html
```diff
@@ -119,6 +119,41 @@
...
+                        <label class="control-label col-md-4 required" for="bddAddIp"
+                               data-i18n-static="privateDatabase_modale_whitelist_add_ip"></label>
+                        <div class="col-md-8">
+                            <input type="text" class="form-control" name="bddAddIp" id="bddAddIp" autocomplete="off" required
+                                   data-ng-model="addBddCtrl.model.ip.value">
+                             <div class="alert alert-info help-block mt-3" id="bddIpHelp" role="alert"
+                                  data-i18n-static="privateDatabase_modale_whitelist_add_ip_tooltip"
+                                  data-ng-class="{'alert-danger': privateDatabaseAddAuthorizedIpForm.bddAddIp.$dirty && !addBddCtrl.isIpValid(addBddCtrl.model.ip.value)}">
```
FredericEspiau: Use Angular form validation functions : `privateDatabaseAddAuthorizedIpForm.bddAddIp.$invalid`

client/app/resources/i18n/privateDatabase/Messages_fr_FR.xml
```diff
@@ -303,6 +303,10 @@
    <translation id="privateDatabase_add_bdd_new_user_password_confirm" qtlid="300539">Confirmer le mot de passe</translation>
    <translation id="privateDatabase_add_bdd_new_user_password_tooltip" qtlid="300552">Entre 8 et 31 caractères, au moins une lettre majuscule, une lettre minuscule et un chiffre, doit être composé uniquement de chiffres et de lettres</translation>
 
+   <translation id="privateDatabase_add_authorized_ip">Ajouter une adresse IP autorisée</translation>
+   <translation id="privateDatabase_add_authorized_ip_explanation">Vous n'avez pas encore d'adresse IP autorisée à accéder à vos bases de données. Entrez ici l'adresse IP du serveur auquel vous souhaitez donner accès à vos bases.</translation>
```
FredericEspiau: This sentence doesn't feel right, wouldn't `Aucune adresse IP n'est autorisée à accéder à vos bases de données. Vous pouvez insérer ici l'adresse IP du ou des serveur(s) qui y auront accès` or something along this way be better ?

(I changed in case more than one server have access to the databases).

client/app/private-database/database/add/private-database-database-add.controller.js
euhmeuh: *addIp* is a boolean linked to the checkbox, in the same style as *addUser*. It does not express if a user can or cannot add an ip, but whether he wants or does not want to add one.
this first initialization means that by default, he wants (we suggest him when the forms opens up).
I'll add a comment to clarify that.

client/app/private-database/database/add/private-database-database-add.html
euhmeuh: It's in the parent's $scope :/

PR #199, opened 2017-11-16
feat(at-internet): update lib and config

whispyy:
- Update ng-at-internet lib to 3.0.0 see https://github.com/ovh-ux/ng-at-internet/pull/4
- Update config file (call to ovhApiMe like in manager Cloud) :
   - customerCode instead of nichandle.
   - countryCode
   - currencyCode

Changes has been already made in Cloud see https://github.com/ovh-ux/ovh-manager-cloud/pull/212


client/app/app.js
```diff
@@ -160,58 +160,19 @@ angular
...
-            return $delegate;
-        });
-    }])
+        OvhApiMe.Lexi().get().$promise
+            .then((me) => {
+                config.countryCode = me.country;
+                config.currencyCode = me.currency && me.currency.code;
+                config.visitorId = me.customerCode;
+                atInternet.setDefaults(config);
+            });
```
jleveugle: no catch ?

whispyy: I'm not able to see what you want me to catch there.
IMO at-internet shouldn't handle a fail of /me.

PR #201, opened 2017-11-16
Feat(domain): update DNS zone deletion process

rfougier:


client/app/domain/zone/delete-all/domain-zone-delete-all.controller.js
```diff
@@ -1,26 +1,44 @@
...
+            this.loading = true;
             this.$scope.deleteAllZone = () => this.deleteAllZone();
+
+            this.checkAllZoneCanBeDelete(this.domain.name);
+        }
+
+        checkAllZoneCanBeDelete (name) {
+            this.loading = true;
+            return this.hostingService
+                .getHosting(name, null)
```
FredericEspiau: Is the second parameter used somewhere else ?

client/app/domain/zone/delete-all/domain-zone-delete-all.controller.js
```diff
@@ -1,26 +1,44 @@
...
+                .then(() => {
+                    this.canDeleteAllZone = false;
+                })
+                .catch((err) => {
+                    this.canDeleteAllZone = err.status === 404;
```
FredericEspiau: I highly doubt you get a `status` property if the service uses `OvhHttp`

rfougier: The service don't

PR #203, opened 2017-11-20
Feat: display database certificate

euhmeuh:
Add a button that allows the user to see and copy/paste his TLS certificate in the database information page.

### DIFF
--- a/client/app/private-database/database/certificate/private-database-database-certificate.controller.js
+++ b/client/app/private-database/database/certificate/private-database-database-certificate.controller.js
@@ -1,0 +1,20 @@
+angular
+    .module("App")
+    .controller("PrivateDatabaseCertificateCtrl", class PrivateDatabaseCertificateCtrl {
+
+        constructor ($scope, $stateParams, Alerter, PrivateDatabase) {
+            this.$scope = $scope;
+            this.$stateParams = $stateParams;
+            this.alerter = Alerter;
+            this.privateDatabaseService = PrivateDatabase;
+        }
+
+        $onInit () {
+            this.productId = this.$stateParams.productId;
+            this.database = this.$scope.currentActionData;
+
+            this.model = {
+                certificate: _.get(this.database, "tlsCa", null)
+            };
+        }
+});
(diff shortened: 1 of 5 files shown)


client/app/private-database/database/certificate/private-database-database-certificate.html
```diff
@@ -0,0 +1,18 @@
+<div data-ng-controller="PrivateDatabaseCertificateCtrl as certificateCtrl">
+
+    <div data-wizard
+         data-wizard-bread-crumb
```
antleblanc: -`data-wizard-bread-crumb`

client/app/private-database/database/certificate/private-database-database-certificate.html
```diff
@@ -0,0 +1,18 @@
...
+         data-wizard-on-cancel="resetAction"
+         data-wizard-cancel-button-text="i18n.privateDatabase_modale_oom_close_button"
+         data-wizard-hide-confirm-button
+         data-wizard-title="i18n.privateDatabase_dashboard_certificate">
+
+        <div data-wizard-step>
+            <div class="form-group">
+                <textarea type="text" rows="10"
+                          class="form-control" id="certificateTextArea"
+                          readonly ng-value="certificateCtrl.database.tlsCa"></textarea>
```
antleblanc: `ng-value` => `data-ng-value` instead?

client/app/private-database/database/certificate/private-database-database-certificate.controller.js
```diff
@@ -0,0 +1,20 @@
...
+            this.alerter = Alerter;
+            this.privateDatabaseService = PrivateDatabase;
+        }
+
+        $onInit () {
+            this.productId = this.$stateParams.productId;
+            this.database = this.$scope.currentActionData;
+
+            this.model = {
+                certificate: _.get(this.database, "tlsCa", null)
```
FredericEspiau: I feel like the difference between `null` and `undefined` shouldn't matter and the code should be handling `null` and `undefined` the same way, thefore `this.database.tlsCa` could suffise, exception of course if `this.database` can be `null` or `undefined`

(I know this is a complexe sentence that I may not have phrased correctly, come to me if you want to have more data ;) )

PR #210, opened 2017-11-23
hotfix(hosting): debounce calls to configRefresh

euhmeuh:


client/app/hosting/hosting.controller.js
```diff
@@ -84,9 +84,10 @@ angular
                                 })
                             );
 
-                            $q.all(queue).then(() => {
-                                loadOvhConfig();
-                            });
+                            $q.all(queue).then(_.debounce(loadOvhConfig, 15000, {
```
jleveugle: don't understand this fix, can you explain why it's necessary please ?

PR #221, opened 2017-12-04
Feat(me-alerts): add new parameter to 2API call

JDownloader:
Add current manager language and universe to 2API me/alerts call. The new version of me-alerts who will soon be deployed will provide some messages already translated.

### Requirements

The code can be merge with or without new version of the 2api in place. Sooner the better, this will avoid modifying each manager if a new message is return by me-alerts.

### Description of the Change

The same modification is done for each manager, the current language of the manager is can change despite /me and subsidiary.


JDownloader:
bower dependency "ovh-api-services" was already to 2.1.9, so changes to bower file should have a minimal impact

PR #225, opened 2017-12-06
hotfix(hosting): remove infinite loading of module list

euhmeuh:


client/app/hosting/module/hosting-module.controller.js
```diff
@@ -50,9 +50,7 @@ angular.module("App").controller(
                     this.Alerter.alertFromSWS(this.$scope.tr("hosting_configuration_tab_modules_create_step1_loading_error"), err, this.$scope.alerts.main);
                 })
                 .finally(() => {
-                    if (_.isEmpty(this.modules.ids)) {
-                        this.loading = false;
-                    }
+                    this.loading = false;
```
rfougier: This test should not be removed.
The loading should be hidden when onTransformItemDone is called.

If you remove this test, loading is removed before API calls are done. The problem must come from another place.

PR #234, opened 2017-12-08
fix(app): redirection to Dedicated now more generic

FredericEspiau:
### Description of the Change

Domains containing "billing" or "useraccount" won't redirect to dedicated anymore

### DIFF
--- a/client/app/app.js
+++ b/client/app/app.js
@@ -219,7 +219,7 @@ angular
             $locationProvider.hashPrefix("");
         }
     ])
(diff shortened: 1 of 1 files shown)


client/app/app.js
```diff
@@ -219,10 +219,12 @@ angular
             $locationProvider.hashPrefix("");
         }
     ])
+    .constant("URLS_REDIRECTED_TO_DEDICATED", [new RegExp("/useraccount/infos"), new RegExp("/billing/history")])
```
jleveugle: It's not only theses 2 urls, we have more urls in billing & user sections

FredericEspiau: @jleveugle you're right, I'm fixing it

PR #240, opened 2017-12-14
Fix(dnsZone): DNS zone ordering works with uppercase domain

rfougier:


client/app/dns-zone/new/dns-zone-new.controller.js
```diff
@@ -22,6 +22,7 @@ angular.module("App").controller(
             this.loading = {
                 bc: false
             };
+            this.zoneName = "";
```
jleveugle: null ? :)

PR #254, opened 2017-12-21
Fix(privateDb): fix creation of database with user

rfougier:


client/app/private-database/database/add/private-database-database-add.controller.js
```diff
@@ -110,8 +110,10 @@ angular.module("App").controller(
                     }
                     return null;
                 })
-                .then(() => this.Alerter.success(this.$scope.tr("privateDatabase_add_bdd_success"), this.$scope.alerts.main),
-                      () => this.Alerter.error(this.$scope.tr("privateDatabase_add_bdd_fail"), this.$scope.alerts.main));
+                .then(() => this.Alerter.success(this.$scope.tr("privateDatabase_add_bdd_success"), this.$scope.alerts.main), (err) => {
+                    _.set(err.data || err, "type", "ERROR");
```
FredericEspiau: This is a terrible idea spread throughout the whole manager, we should fix the root cause, not the consequences

rfougier: So we need to fix the API...

PR #268, opened 2018-01-05
fix(domain): allow DS Records modification, limit their number to 4

euhmeuh:


client/app/domain/dnssec/domain-dnssec.controller.js
```diff
@@ -92,11 +92,17 @@ angular.module("controllers").controller(
         }
 
         addRecord () {
-            this.dnssecList.push({ id: this.dnssecList.length + 1, tag: 0, flags: "", algorithm: "", publicKey: "" });
+            this.dnssecList.push({ id: this.nextAvailableId(), tag: 0, flags: "", algorithm: "", publicKey: "" });
```
jleveugle: Can we avoid empty string ? Can we use `null` value ?

client/app/domain/dnssec/domain-dnssec.controller.js
```diff
@@ -92,11 +92,17 @@ angular.module("controllers").controller(
...
+        }
+
+        nextAvailableId () {
+            return _.find(_.range(this.const.MAX_AMOUNT_DNSSEC),
+                          (id) => !_.any(this.dnssecList,
```
jleveugle: `_.any(this.dnssecList, { id })`

client/app/domain/domain.service.js
```diff
@@ -1096,7 +1096,7 @@ angular.module("services").service(
         getDomainPendingTasks (serviceName, data) {
             return this.OvhHttp.get(`/domain/${serviceName}/task`, {
                 rootPath: "apiv6",
-                data
+                params: data
```
jleveugle: You should rename data by params to get benefits of ES6 syntax :)

PR #273, opened 2018-01-11
Fix (translations) core: 402074 & 402087

janphilippi:
### Description of the Change

Remove unnecessary characters from qtlid 402074 & 402087


jleveugle:
The language translations sentences must be done in French only 🇫🇷  (temporarily, due to internal process).

Please see https://github.com/ovh-ux/ovh-manager-web/blob/master/.github/CONTRIBUTING.md

PR #276, opened 2018-01-16
hotfix(hosting/database): filter out mongodb and redis

euhmeuh:
Redis and mongodb were added to the backend, but not released yet.
I'm hiding them until we prod.

The long function name is by design so that one does immediately know it's a quick temp fix.

### DIFF
--- a/client/app/hosting/database/private-activation/database-private-activation.controller.js
+++ b/client/app/hosting/database/private-activation/database-private-activation.controller.js
@@ -33,7 +33,7 @@ angular.module("App").controller("HostingDatabasePrivateActiveCtrl", ($scope, $r
 
     function getAvailableVersions (models) {
         const types = models.models["hosting.PrivateDatabase.OrderableVersionEnum"].enum;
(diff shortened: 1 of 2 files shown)


client/app/hosting/database/private-activation/database-private-activation.controller.js
```diff
@@ -19,8 +19,8 @@ angular.module("App").controller("HostingDatabasePrivateActiveCtrl", ($scope, $r
             $scope.loaders.versions = true;
             Hosting.getModels()
                 .then((models) => {
-                    $scope.datas.versions = models.models["hosting.PrivateDatabase.OrderableVersionEnum"].enum;
-                    $scope.choice.version = $scope.datas.versions.length === 1 ? $scope.datas.versions[0] : null;
+                    $scope.data.versions = getAvailableVersions();
+                    $scope.choice.version = $scope.data.versions.length === 1 ? $scope.data.versions[0] : null;
```
FredericEspiau: `_($scope.data.versions).isArray()`

euhmeuh:
@jleveugle I agree. We do that because the API returns not-yet-prodded models. I want to replace that with a correct call to the right API endpoint, but right now we need to lower the support calls from angry customers. :(

PR #277, opened 2018-01-16
fix(domain.zoneDns) remove the inline add dns zone form

cbourgois:
### Description of the Change
Remove inline form to add new DNS record


antleblanc:
@cbourgois please PR to the `develop` branch.

PR #278, opened 2018-01-18
Feature/oui datagrid hosting

JDownloader:
5 tables were converted to datagrids

Modules   -> /#/configuration/hosting/<serviceName>?tab=MODULE
Bases de données - Dumps   -> /#/configuration/hosting/<serviceName>?tab=DATABASE
Tâches en cours   -> /#/configuration/hosting/<serviceName>?tab=TASK
Statistiques et Logs (user-logs)   -> /#/configuration/hosting/<serviceName>?tab=USER_LOGS
Boost   -> /#/configuration/hosting/<serviceName>?tab=BOOST

### Before reviewing, need this PR to be merged first !
https://github.com/ovh-ux/ovh-manager-web/pull/267


client/app/hosting/user-logs/hosting-user-logs.controller.js
```diff
@@ -70,32 +65,34 @@ angular.module("App").controller(
         }
 
         refreshTableUserLogs () {
-            this.loaders.userLogs = true;
             this.userLogs.ids = null;
 
             return this.Hosting.getUserLogs(this.$stateParams.productId)
-                .then((data) => {
-                    this.userLogs.ids = this.$filter("orderBy")(data);
+                .then((data) => data.sort())
```
FredericEspiau: you should check that data is indeed an array before using the `sort` method. You should also be aware of the shortcomings of using `sort`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort

Isn't Lodash more appropriate ?

client/app/hosting/user-logs/USER_LOGS.html
```diff
@@ -44,78 +44,29 @@ <h2 data-i18n-static="hosting_tab_USER_LOGS"></h2>
...
+                        <span data-ng-bind="$row.login"></span>
+                    </oui-column>
+                    <oui-column data-title="tr('hosting_tab_USER_LOGS_table_header_description')" data-property="description">
+                        <span data-ng-bind="$row.description | date:'short'"></span>
+                    </oui-column>
+                    <oui-column data-title="tr('hosting_tab_USER_LOGS_table_header_creationDate')" data-property="creationDate">
+                        <span data-ng-bind="$row.creationDate | date: 'medium'"></span>
+                    </oui-column>
+                    <oui-action-menu align="end" compact>
+                        <oui-action-menu-item text="{{::tr('hosting_tab_USER_LOGS_table_popover_modify')}}"
```
FredericEspiau: Weren't you supposed to add `data-` to all custom attributes ?

client/app/hosting/task/hosting-task.controller.js
```diff
@@ -17,8 +17,7 @@ angular.module("App").controller(
                 TODO: "TODO",
                 INIT: "INIT"
             };
-            this.tasksList = null;
-            this.loading = false;
+            this.tasksList = undefined;
```
FredericEspiau: Why would you define to `undefined` by default, this is already the default value

client/app/hosting/database/hosting-database.service.js
```diff
@@ -92,14 +92,33 @@ angular
...
+                return dump;
             });
         }
 
+        /**
+         * Get snapshot date of database dump
+         * @param {Object} dump
+         */
+        static getSnapshotDateOfDump (dump) {
+            if (!dump.creationDate) { return undefined; }
```
FredericEspiau: ESLint shouldn't allow this, this is not the way we indent code :). Didn't you have warnings ?

FredericEspiau: I'd also rather have you check if `dump.creationDate` is a string and is not empty:
`_(dump.creationDate).isString() && !_(dump.creationDate).isEmpty()`

client/app/hosting/database/dump/hosting-database-dump.controller.js
```diff
@@ -123,7 +106,7 @@ angular.module("App").controller(
 
             if (dump && dump.id) {
```
FredericEspiau: `_(dump).isObject() && _(dump).isFinite(dump.id)`

client/app/hosting/database/dump/hosting-database-dump.controller.js
```diff
@@ -24,52 +22,37 @@ angular.module("App").controller(
         }
 
         loadDumps () {
-            this.dumpsLoading = true;
-            this.dumpsIds = null;
+            this.databaseDumps = undefined;
```
FredericEspiau: `undefined` is already the default value, you don't have to write this

client/app/hosting/user-logs/USER_LOGS.html
JDownloader: Yes, will fix

client/app/hosting/user-logs/hosting-user-logs.controller.js
JDownloader: I would have preferred an argument like «sorting could be done by the service», If the Api is giving me a 200 without being an Array, screw it, I don't care. 

I think  ```Cannot read property 'sort' of undefined``` would be totaly fine. What do you want me to do, throw an exception if it's non an array? I don't have to write code for that.

client/app/hosting/task/hosting-task.controller.js
JDownloader: I had almost the same argument with another developper on not assigning undefined, this is only for code readability

client/app/hosting/user-logs/hosting-user-logs.controller.js
FredericEspiau: Well in this case you're showing the user the error `Cannot read property 'sort' of undefined` instead of showing an error message that can be translated into his language :). But I see what you mean, I've been doing the same quite a few time.

client/app/hosting/database/dump/hosting-database-dump.controller.js
JDownloader: first, ```if (_(dump.get(id))``` would be a lot simpler
second, this is not part of this pull request
third, lowdash is making our code bloated

client/app/hosting/user-logs/USER_LOGS.html
FredericEspiau: Neat :)

client/app/hosting/database/dump/hosting-database-dump.controller.js
JDownloader: this one ensure datagrid is loading while refreshing

client/app/hosting/database/hosting-database.service.js
JDownloader: One-liner are completely fine, since it could be seen a single instruction.

THERE'S NO WARNING, please pull my code and try on your computer before make some bold assumptions. Look at the test phase with our CI, no problem, only success.

This pull request is in accordance with the project ESLint config

JDownloader: I could also check every String return by the Api, move on and stop making me using lodash for everything

antleblanc:
I noticed that you are doing Pull Request to an another branch name (here: `chore/update-oui-kit`) that meaning we eventually can squash commits at the end of the milestone, but I really encourage you to stick with some commit name conventions. Then please could you update your commit messages?

Here you can find all needed informations: https://github.com/ovh-ux/ovh-ux-guidelines/blob/master/.github/CONTRIBUTING.md#commit-message-format.

client/app/hosting/user-logs/hosting-user-logs.controller.js
Mboulianne: @FredericEspiau  out of curiosity, how do you translate "a dev screwed up" into user language?  We could say an error occured while fetching ZYX which leads the user to think he can try to reload or something, but isn't the case.

JDownloader: I forget to tell that in this specific case there's no «shortcomings», since it's a list of unix user with or accents or special characters. 

On that assertion problem, since javascript is an Object Language with error, I think we must escape that 90s trend of [defensive software](https://en.wikipedia.org/wiki/Defensive_programming). I use those when that kind of writing style only when I know that the API can screw-up.

AxelPeter:
@JDownloader Careful on your commit messages. See @antleblanc comment.

@antleblanc the target branch will be develop, but it need to be merged first :)

AxelPeter:
https://github.com/ovh-ux/ovh-manager-web/pull/267 has been merged. I change back the target branch to `develop`

client/app/hosting/boost/BOOST.html
```diff
@@ -19,71 +19,27 @@ <h2 data-i18n-static="hosting_tab_BOOST"></h2>
...
+            <oui-datagrid data-rows="hostingBoosts"
+                          data-row-loader="transformItem($row)">
+                <oui-column data-title="tr('hosting_tab_BOOST_history_table_header_id')" data-property="accountId">
+                    <span data-ng-bind="$row.accountId"></span>
+                </oui-column>
+                <oui-column data-title="tr('hosting_tab_BOOST_history_table_header_offer')" data-property="offer">
+                    <span data-ng-bind="$row.offer"></span>
+                </oui-column>
+                <oui-column data-title="tr('hosting_tab_BOOST_history_table_header_boost_offer')" data-property="boostOffer">
+                    <span data-ng-class="{center: $row.boostOffer === null}">
```
jleveugle: text-center :)

center class shouldn't exist :p

JDownloader: Why does this class exist then? I just replicated what was there before. :-|

PR #281, opened 2018-01-19
Show the right database backends available

euhmeuh:
We were not using the right API endpoint to know what were the database types available to purchase when going to Hosting (perf only) > Private SQL > Activate.

I refactored the controller to a class and hit the right endpoint.
Now every time the API staff pushes changes or new database backends, they are available to purchase.

(don't mind the branch name, Redis is prodded now)

### DIFF
--- a/client/app/hosting/database/private-activation/database-private-activation.controller.js
+++ b/client/app/hosting/database/private-activation/database-private-activation.controller.js
@@ -33,7 +33,7 @@ angular.module("App").controller("HostingDatabasePrivateActiveCtrl", ($scope, $r
 
     function getAvailableVersions (models) {
         const types = models.models["hosting.PrivateDatabase.OrderableVersionEnum"].enum;
-        return temporarilyFilterTypesUntilTheyAreOfficiallyReleased(types, ["mongodb_3.4", "redis_3.2"]);
+        return temporarilyFilterTypesUntilTheyAreOfficiallyReleased(types, ["mongodb_3.4"]);
     }
 
     function temporarilyFilterTypesUntilTheyAreOfficiallyReleased (types, restrictedList) {
--- a/client/app/hosting/database/private-activation/database-private-activation.controller.js
+++ b/client/app/hosting/database/private-activation/database-private-activation.controller.js
@@ -3,7 +3,7 @@ angular.module("App").controller("HostingDatabasePrivateActiveCtrl", ($scope, $r
 
     $scope.hosting = $scope.currentActionData;
 
-    $scope.datas = {
+    $scope.data = {
         versions: []
     };
     $scope.loaders = {
@@ -19,8 +19,8 @@ angular.module("App").controller("HostingDatabasePrivateActiveCtrl", ($scope, $r
             $scope.loaders.versions = true;
             Hosting.getModels()
                 .then((models) => {
-                    $scope.datas.versions = models.models["hosting.PrivateDatabase.OrderableVersionEnum"].enum;
-                    $scope.choice.version = $scope.datas.versions.length === 1 ? $scope.datas.versions[0] : null;
+                    $scope.data.versions = getAvailableVersions(models);
+                    $scope.choice.version = $scope.data.versions.length === 1 ? $scope.data.versions[0] : null;
                 })
                 .catch((err) => {
                     Alerter.alertFromSWS($scope.tr("hosting_dashboard_database_versions_error", [$scope.entryToDelete]), _.get(err, "data", err), $scope.alerts.main);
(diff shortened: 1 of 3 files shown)


client/app/hosting/database/private-activation/database-private-activation.controller.js
```diff
@@ -1,52 +1,71 @@
...
-        versions: []
-    };
-    $scope.loaders = {
-        versions: false
-    };
-    $scope.choice = {
-        ram: $scope.hosting.offerCapabilities.privateDatabases.length === 1 ? $scope.hosting.offerCapabilities.privateDatabases[0] : null,
-        version: null
-    };
+            this.host = this.$scope.currentActionData;
```
FredericEspiau: This and all the lines below should be included in the `$onInit` function

client/app/hosting/database/private-activation/database-private-activation.controller.js
```diff
@@ -1,52 +1,71 @@
...
+            this.versions = [];
+            this.loaders = {
+                versions: false
```
FredericEspiau: I guess this is related to data about versions being retrieved from the API so this should be called `versionDataRetrieval` or something of this type

client/app/hosting/database/private-activation/database-private-activation.controller.js
```diff
@@ -1,52 +1,71 @@
...
+            };
+            this.choice = {
+                ram: this.host.offerCapabilities.privateDatabases.length === 1 ?
```
FredericEspiau: I don't understand this kind of thing : only have a default choice only if there is only 1 element ? Does it really make sense ? If the user doesn't have a choice, then we shouldn't display a select box.

client/app/hosting/database/private-activation/database-private-activation.controller.js
```diff
@@ -1,52 +1,71 @@
...
+            this.$scope.activateDatabase = () => this.activateDatabase();
         }
-    }
 
-    $scope.activeDatabase = function () {
-        if (!$scope.loaders.versions) {
-            $scope.loaders.versions = true;
-            HostingDatabase.activeDatabasePrivate($stateParams.productId, $scope.choice.ram.quota.value, $scope.choice.version)
+        $onInit () {
+            if (!this.loaders.versions) {
```
FredericEspiau: How could `this.loaders.version` be `true` ?

client/app/hosting/database/private-activation/database-private-activation.controller.js
```diff
@@ -1,52 +1,71 @@
...
+                this.loaders.versions = true;
+                this.PrivateDatabase.getAvailableOrderCapacities("classic")
+                    .then((capacities) => {
+                        this.versions = _.get(capacities, "version");
```
FredericEspiau: You're trying to prevent a `TypeError` by not getting `version` if it doesn't exist and then you call `length` on `this.version`, fully expecting `this.versions` to be an array.  Please check the types more carefully :)

client/app/hosting/database/private-activation/database-private-activation.controller.js
```diff
@@ -1,52 +1,71 @@
...
+            }
+        }
+
+        activateDatabase () {
+            if (this.loaders.versions) {
+                return;
+            }
+
+            this.loaders.versions = true;
+            this.HostingDatabase.activateDatabasePrivate(this.$stateParams.productId,
```
FredericEspiau: `return`

client/app/hosting/database/private-activation/database-private-activation.controller.js
```diff
@@ -1,52 +1,71 @@
...
+                                                         this.choice.ram.quota.value,
+                                                         this.choice.version)
                 .then(() => {
-                    $rootScope.$broadcast("hosting.database.sqlPrive");
-                    Alerter.success($scope.tr("hosting_dashboard_database_active_success"), $scope.alerts.main);
+                    this.$rootScope.$broadcast("hosting.database.sqlPrive");
```
FredericEspiau: Well this is not a lot explicit, what does it do ?

client/app/hosting/database/private-activation/database-private-activation.controller.js
```diff
@@ -1,52 +1,71 @@
...
+                                                  this.$scope.alerts.main);
+                    })
+                    .finally(() => {
+                        this.loaders.versions = false;
+                    });
+            }
+        }
+
+        activateDatabase () {
+            if (this.loaders.versions) {
```
FredericEspiau: This shouldn't be possible now, should it ?

client/app/hosting/database/private-activation/database-private-activation.controller.js
euhmeuh: You're right, I failed to notice this useless if.

PR #313, opened 2018-02-14
fix(modal): update backdrop color for modals

neolitec:
UU-716

### DIFF
--- a/client/app/css/source.less
+++ b/client/app/css/source.less
@@ -47,3 +47,5 @@
 
 @ovh-font-path: "../bower_components/ovh-manager-webfont/dist/fonts";
 @ovh-universe-color: @oui-color-jungle;
+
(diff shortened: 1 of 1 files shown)


client/app/css/source.less
```diff
@@ -47,3 +47,5 @@
 
 @ovh-font-path: "../bower_components/ovh-manager-webfont/dist/fonts";
 @ovh-universe-color: @oui-color-jungle;
+
+@modal-backdrop-bg: @oui-color-shuttle;
```
AxelPeter: Should be in https://github.com/ovh-ux/ovh-ui-kit-bs

AxelPeter:
The fix is on ovh-ui-kit-bs@v1.0.5 : https://github.com/ovh-ux/ovh-ui-kit-bs/releases/tag/v1.0.5
I think we can close this PR :)

PR #315, opened 2018-02-20
feat(navbar): add hierarchical user menu

AxelPeter:
UU-735


AxelPeter:
@whispyy 
Versions have changed in the `package.json`, why do we don't have to pushed the `package-lock.json` ?

AxelPeter:
I remove the changes for the responsive bug. It will be done in another PR because it need this one to be merged first: https://github.com/ovh-ux/ovh-angular-sidebar-menu/pull/18.

client/app/index.ejs
```diff
@@ -90,7 +91,8 @@
             </div>
         </div>
 
-        <div id="managerPreload" class="d-flex text-center" data-ng-class="managerPreloadHide">
+        <div id="managerPreload" class="d-flex text-center"
+            data-ng-class="!!navbar && managerPreloadHide">
```
FredericEspiau: Using the class in the variable `managerPreloadHide` only if `navbar` is defined ? Wouldn't it be clearer in a function ?

client/app/components/user/session.service.js
```diff
@@ -241,25 +245,114 @@ class SessionService {
...
-                }, {
+                    url: "#/useraccount/infos",
+                    subLinks: [{
+                        title: this.translator.tr("global_account_infos"),
+                        url: "#/useraccount/infos"
+                    }, {
+                        title: this.translator.tr("global_account_security"),
+                        url: "#/useraccount/security"
+                    },
+                        (this.constants.target === "EU" || this.constants.target === "CA") && {
```
FredericEspiau: I'm conflicted as I like the brevity of it but I also find it quite hard to read.

Wouldn't this be clearer if it were behind a `if` statement ? Or maybe a function that adds the items only if a condition is fulfilled ?

`function addItem(title, url, condition)` ? I don't know, just a thought

AxelPeter: Yes, it was to avoid the combination of `if` and `push`, since we have condition for the subLinks.
Your idea is to replace it with something like
```
addItem(!currentUser.isEnterprise, {
    name: "user.billing",
    title: this.translator.tr("global_billing"),
    url: "#/billing/history",
    subLinks: [{
        title: this.translator.tr("global_billing_history"),
        url: "#/billing/history"
    }, {
        title: this.translator.tr("global_billing_payments"),
        url: "#/billing/payments"
    }]
})
```
?

client/app/index.ejs
AxelPeter: Sadly, it's like that in every manager. In Web, the classname is added on `$stateChangeSuccess`.
Here I just check that the navbar is defined before adding the class, to avoid the possibility of a blank space.

client/app/components/user/session.service.js
euhmeuh: IMHO I like it that way. Won't it insert false values in the structure though?

AxelPeter: Yes, it's for the one-time binding in the navbar component. It keep a watcher since the value is not anymore `undefined`. The `false` value will be ignored when building the menu.

PR #316, opened 2018-02-21
style(*): update layout responsive

AxelPeter:
Avoid the Sidebar and the responsive menu of the navbar to be displayed at the same time.
The breakpoint is at 1024px.

**Will need this PR to be merged first https://github.com/ovh-ux/ovh-angular-sidebar-menu/pull/18**

### DIFF
--- a/client/app/components/sidebar-menu/sidebar-menu.less
+++ b/client/app/components/sidebar-menu/sidebar-menu.less
@@ -10,8 +10,6 @@ button, toggle-class {
 }
 
 #sidebar-menu {
-    z-index: 1020;
-
     .ms-Icon {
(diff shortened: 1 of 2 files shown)


client/app/components/sidebar-menu/sidebar-menu.less
```diff
@@ -10,8 +10,6 @@ button, toggle-class {
 }
 
 #sidebar-menu {
-    z-index: 1020;
```
FredericEspiau: This is beautiful

AxelPeter: No need to override it now :D

PR #124, opened 2017-10-24
feat(privateDb): Change display to show max quota for private DB

Mboulianne:
### Requirements

* Filling out the template is required. Any pull request that does not include enough information to be reviewed in a timely manner may be closed at the maintainers' discretion.
* All new code requires tests to ensure against regressions

PR #321, opened 2018-02-26
fix(email-domain): Fix for using deprecated API call

miecio45:
### Description of the Change
Remove all calls to deprecated APIv6 call for imapCopy, as it is now
replaced by OVH Mail Migrator (OMM), I added OMM link to the side panel.
I used translation from old ImapCopy just to not force translation again.

### Benefits
No deprecated API call will be used, link to OMM is present in manager

### Possible Drawbacks
Need to check that I did not miss anything

### Applicable Issues
Nothing for now


FredericEspiau:
Nice work

miecio45:
I removed all translation for this modal window

client/app/email-domain/email/email-domain-email.controller.js
```diff
@@ -123,6 +125,11 @@ angular
             this.$window.open(this.webMailUrl, "_blank");
         }
 
+
+        openOMMTab () {
```
jleveugle: Can we use a link instead of button + ngClick + $window.open ?

It's just a `<a href="..." target="_blank"></a>` ... Don't forget title attribute title attribute for accessibility if needed :)

miecio45: I just followed example from line [ #124 ](https://github.com/miecio45/ovh-manager-web/blob/0d8a8d9afa9ee1f0ce392b81e6931521b329d34a/client/app/email-domain/email/email-domain-email.controller.js#L124) for webmail page link and assume that this is the model solution

jleveugle: Yes we know our code and its weak points.

Currently, we are improving our source code to improve the product. I invite you to follow my recommendations.

miecio45: I hope that I fix it :)

PR #328, opened 2018-02-28
hotfix(sidebar-menu): remove race condition when loading …

euhmeuh:
There was a rare bug when loading the page on an account with few products.
The promise was finishing too fast, leading to a race condition with the "Order" button loading.

### DIFF
--- a/client/app/components/sidebar-menu/sidebar-menu.config.js
+++ b/client/app/components/sidebar-menu/sidebar-menu.config.js
@@ -3,6 +3,75 @@ angular.module("App").run(($q, SidebarMenu, Products, User, constants, translato
 
     const menuOptions = [];
 
+    function buildMenuOptions () {
+        return User.getUrlOf("domainOrder")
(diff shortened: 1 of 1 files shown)


client/app/components/sidebar-menu/sidebar-menu.config.js
```diff
@@ -3,6 +3,75 @@ angular.module("App").run(($q, SidebarMenu, Products, User, constants, translato
 
     const menuOptions = [];
 
+    function buildMenuOptions () {
+        return User.getUrlOf("domainOrder")
```
cbourgois: Avoid chaining `then`
```
return $q.all([User.getUrlOf("domainOrder"), User.getUrlOf("emailproOrder"), User.getUrlOf("office365Order")])
    .then(([domainOrderUrl, emailproOrderUrl, office365OrderUrl]) => { 
        // menuOptions.push calls
    });
```

PR #335, opened 2018-03-02
feat(chore): bump ui-router to version ^1.0.0.

frenautvh:
bump ui-router to version 1.0.0
with updated dependencies


frenautvh:
@jleveugle do you think we can merge this one ?

client/app/app.js
```diff
@@ -472,66 +472,67 @@ angular
...
-                            }
-                            if (value.translations) {
-                                translationsTab = _.union(translationsTab, value.translations);
-                            }
-                        }
-
-                    });
-                }
+                        if (state.views) {
+                            angular.forEach(state.views, (value) => {
```
FredericEspiau: Why use `angular`, we mostly use `_`

FredericEspiau: It looks like you could have two `map` instead of one `foreach`, maybe it would be more readable

client/app/app.js
```diff
@@ -472,66 +472,67 @@ angular
...
-                return translationsTab;
-            }
+                        return $translate.refresh();
+                    }
 
-            return null;
-        });
+                    return null;
+                }
+            }); // translation.addResolvable
```
FredericEspiau: huh, this should not be needed, maybe you could extract some code to function so the `addResolvable` is shorter

client/app/app.js
```diff
@@ -544,11 +545,11 @@ angular
         partners: "https://www.ovh.com/manager/partners/",
         labs: "https://www.ovh.com/manager/sunrise/uxlabs/#!/"
     })
-    .run(($rootScope, translator,
+    .run(($rootScope, $transitions, translator,
           ouiDatagridConfiguration, ouiPaginationConfiguration) => {
         "use strict";
 
-        const off = $rootScope.$on("$stateChangeSuccess", () => {
+        const removeHook = $transitions.onSuccess({}, () => {
```
FredericEspiau: What hook ? You could name it `removeOnSuccessHook` to be more specific

antleblanc:
@jleveugle conflicts are now fixed (rebase on `develop` branch).

PR #347, opened 2018-03-13
fix(email.delegate): Set property to display quota progress bar

marie-j:
### Description of the Change

Before: There was only a textual display of the quota
Now: A progress bar is displayed along with the textual display


client/app/email-domain/delegate/email-domain-delegate.controller.js
```diff
@@ -114,6 +114,13 @@ angular.module("App").controller(
                     email.quota = usage.quota;
                     email.emailCount = usage.emailCount;
                     email.date = usage.date;
+
+                    if (email.size > 0) {
```
FredericEspiau: Apparently this code is already present in the controller, maybe you would extract this bit of code in a function so you don't repeat yourself : https://en.wikipedia.org/wiki/Don%27t_repeat_yourself :smiley:

PR #357, opened 2018-03-16
fix(domain.zone) : Limit form submission

cbourgois:
### Description of the Change

Limit DNS zone form submission to avoid multiples API calls


FredericEspiau:
This only fixes 1 dialog, modifying the Wizard component would fix all of them

jleveugle:
Quality check ok ? :)

PR #366, opened 2018-03-22
Chore/privatedb update database minimum length

marie-j:
### Description of the Change

Before: The database name had to have at least 3 characters
Now: The database name minimum length is 1 character

### DIFF
--- a/client/app/private-database/database/add/private-database-database-add.controller.js
+++ b/client/app/private-database/database/add/private-database-database-add.controller.js
@@ -33,10 +33,10 @@ angular.module("App").controller(
                     value: "",
                     condition: {
                         length: {
-                            min: 3,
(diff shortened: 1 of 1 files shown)


client/app/private-database/database/add/private-database-database-add.controller.js
```diff
@@ -33,10 +33,10 @@ angular.module("App").controller(
                     value: "",
                     condition: {
                         length: {
-                            min: 3,
+                            min: 1,
```
jleveugle: Can we get the information from API ?

marie-j: I had a look and API doesn't give that kind of information unfortunately

PR #371, opened 2018-03-23
chore(private_db): Upgrade mysql configuration for ui-kit

JDownloader:
Project uniformisation

### DIFF
--- a/client/app/private-database/configuration/private-database-configuration.controller.js
+++ b/client/app/private-database/configuration/private-database-configuration.controller.js
@@ -2,10 +2,11 @@
 angular.module("App").controller(
     "PrivateDatabaseConfigurationsCtrl",
     class PrivateDatabaseConfigurationsCtrl {
-        constructor (Alerter, PrivateDatabase, translator, $scope, $stateParams) {
+        constructor (Alerter, PrivateDatabase, translator, $q, $scope, $stateParams) {
             this.alerter = Alerter;
             this.privateDatabaseService = PrivateDatabase;
             this.translator = translator;
+            this.$q = $q;
             this.$scope = $scope;
             this.$stateParams = $stateParams;
         }
@@ -29,8 +30,9 @@ angular.module("App").controller(
                 .then((config) => {
                     this.configurations = config.details.map((field) => this.convertAvailableValues(field));
                 })
-                .catch(() => {
+                .catch((error) => {
                     this.alerter.error(this.translator.tr("privateDatabase_configuration_error"), this.$scope.alerts.main);
+                    this.$q.reject(error);
                 })
                 .finally(() => {
                     this.loading = false;
@@ -60,18 +62,21 @@ angular.module("App").controller(
         getFieldDescriptionTranslated (field) {
             const translationId = `privateDatabase_configuration_field_${field.key}`;
             let description = this.translator.tr(translationId);
-            if (description.includes(translationId)) {
+            if (_.includes(description, translationId)) {
                 description = field.description;
             }
             return description;
         }
 
         updateConfigurations () {
+            if (this.dbConfigurationForm.$invalid) {
+                return this.$q.reject();
+            }
             this.loading = true;
             this.edit.value = false;
             const parameters = this.configurations.map((conf) => ({ key: conf.key, value: conf.selectedValue.id }));
 
-            this.privateDatabaseService
+            return this.privateDatabaseService
                 .changeConfigurationDetails(this.productId, { parameters })
                 .then(() => {
                     this.alerter.success(this.translator.tr("privateDatabase_configuration_reboot"), this.$scope.alerts.main);
@@ -80,8 +85,9 @@ angular.module("App").controller(
                 .then(() => {
                     this.alerter.success(this.translator.tr("privateDatabase_configuration_success"), this.$scope.alerts.main);
                 })
-                .catch(() => {
+                .catch((error) => {
                     this.alerter.error(this.translator.tr("privateDatabase_configuration_error_put"), this.$scope.alerts.main);
+                    this.$q.reject(error);
                 })
                 .finally(() => {
                     this.edit.value = false;
--- a/client/app/private-database/configuration/private-database-configuration.controller.js
+++ b/client/app/private-database/configuration/private-database-configuration.controller.js
@@ -2,9 +2,10 @@
 angular.module("App").controller(
     "PrivateDatabaseConfigurationsCtrl",
     class PrivateDatabaseConfigurationsCtrl {
-        constructor (Alerter, PrivateDatabase, $scope, $stateParams) {
+        constructor (Alerter, PrivateDatabase, translator, $scope, $stateParams) {
             this.alerter = Alerter;
             this.privateDatabaseService = PrivateDatabase;
+            this.translator = translator;
             this.$scope = $scope;
             this.$stateParams = $stateParams;
         }
@@ -14,7 +15,6 @@ angular.module("App").controller(
 
             this.database = this.$scope.database;
             this.loading = false;
-            this.isRebooting = false;
             this.edit = {
                 value: false
             };
@@ -24,41 +24,68 @@ angular.module("App").controller(
 
         getConfigurationDetails () {
             this.loading = true;
-            this.privateDatabaseService
+            return this.privateDatabaseService
                 .getConfigurationDetails(this.productId)
                 .then((config) => {
-                    this.configurations = config.details;
-                    this.baseConfiguration = angular.copy(config.details);
+                    this.configurations = config.details.map((field) => this.convertAvailableValues(field));
                 })
-                .catch((err) => {
-                    this.alerter.alertFromSWS(this.$scope.tr("privateDatabase_configuration_error"), err, this.$scope.alerts);
+                .catch(() => {
+                    this.alerter.error(this.translator.tr("privateDatabase_configuration_error"), this.$scope.alerts.main);
                 })
                 .finally(() => {
                     this.loading = false;
                 });
         }
 
+        convertAvailableValues (field) {
+            field.description = this.getFieldDescriptionTranslated(field);
+            field.availableValues = field.availableValues.map((value) => ({ id: value, text: value }));
+            if (field.key === "autocommit") {
+                PrivateDatabaseConfigurationsCtrl.convertFieldAsToggle(field);
+                field.availableValues = [{ id: "0", text: " OFF" }, { id: "1", text: "ON" }];
+            } else if (field.key === "event_scheduler") {
+                PrivateDatabaseConfigurationsCtrl.convertFieldAsToggle(field);
+            } else {
+                field.type = "select";
+                field.selectedValue = field.availableValues.find((value) => value.id === field.value);
+            }
+            return field;
+        }
+
+        static convertFieldAsToggle (field) {
+            field.type = "toggle";
+            field.selectedValue = { id: field.value };
+        }
+
+        getFieldDescriptionTranslated (field) {
+            const translationId = `privateDatabase_configuration_field_${field.key}`;
+            let description = this.translator.tr(translationId);
+            if (description.includes(translationId)) {
+                description = field.description;
+            }
+            return description;
+        }
+
         updateConfigurations () {
             this.loading = true;
             this.edit.value = false;
-            const parameters = _.map(this.configurations, (conf) => ({ key: conf.key, value: conf.value }));
+            const parameters = this.configurations.map((conf) => ({ key: conf.key, value: conf.selectedValue.id }));
 
             this.privateDatabaseService
                 .changeConfigurationDetails(this.productId, { parameters })
                 .then(() => {
-                    this.alerter.success(this.$scope.tr("privateDatabase_configuration_reboot"), this.$scope.alerts);
+                    this.alerter.success(this.translator.tr("privateDatabase_configuration_reboot"), this.$scope.alerts.main);
                     return this.privateDatabaseService.pollConfigurationChange(this.productId);
                 })
                 .then(() => {
-                    this.alerter.success(this.$scope.tr("privateDatabase_configuration_success"), this.$scope.alerts);
+                    this.alerter.success(this.translator.tr("privateDatabase_configuration_success"), this.$scope.alerts.main);
                 })
-                .catch((err) => {
-                    this.alerter.alertFromSWS(this.$scope.tr("privateDatabase_configuration_error_put"), err, this.$scope.alerts);
+                .catch(() => {
+                    this.alerter.error(this.translator.tr("privateDatabase_configuration_error_put"), this.$scope.alerts.main);
                 })
                 .finally(() => {
                     this.edit.value = false;
                     this.loading = false;
-                    this.isRebooting = false;
                     this.getConfigurationDetails();
                 });
         }
--- a/client/app/private-database/configuration/private-database-configuration.html
+++ b/client/app/private-database/configuration/private-database-configuration.html
@@ -1,76 +1,67 @@
 <div class="container-fluid px-0" data-ng-controller="PrivateDatabaseConfigurationsCtrl as configurationsCtrl">
 
+    <div data-ovh-alert="{{alerts.main}}"></div>
+
     <div class="text-center" data-ng-if="configurationsCtrl.loading">
         <oui-spinner data-size="l"></oui-spinner>
     </div>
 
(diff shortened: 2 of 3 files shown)


client/app/private-database/configuration/private-database-configuration.html
```diff
@@ -1,76 +1,67 @@
...
     <div class="text-center" data-ng-if="configurationsCtrl.loading">
         <oui-spinner data-size="l"></oui-spinner>
     </div>
 
-    <div class="row" data-ng-if="!configurationsCtrl.loading">
-        <div class="col-md-9">
-            <div data-ovh-alert="{{alerts.main}}"></div>
+    <div data-ng-if="!configurationsCtrl.loading">
+
+        <div class="alert alert-info" role="alert"
```
tremblaymath: No bootstrap alert, only ui-kit.
You should use oui-message

client/app/private-database/configuration/private-database-configuration.html
```diff
@@ -1,76 +1,67 @@
...
+             data-i18n-static="privateDatabase_configurations_general_postgresql_soon"
+             data-ng-if="configurationsCtrl.database.type === 'postgresql'">
+        </div>
+        <div class="alert alert-info" role="alert"
```
tremblaymath: Same

client/app/private-database/configuration/private-database-configuration.html
```diff
@@ -1,76 +1,67 @@
...
+             data-i18n-static="privateDatabase_capability_disabled_explanation"
+             data-ng-if="configurationsCtrl.database.type !== 'postgresql' && configurationsCtrl.configurations.length === 0">
+        </div>
+
+        <div data-ng-if="configurationsCtrl.configurations.length >= 1">
 
-            <div class="alert alert-info" role="alert"
-                 data-i18n-static="privateDatabase_configurations_general_postgresql_soon"
-                 data-ng-if="configurationsCtrl.database.type === 'postgresql'">
+            <div class="alert alert-warning"
```
tremblaymath: Same

client/app/private-database/configuration/private-database-configuration.html
```diff
@@ -1,76 +1,67 @@
 <div class="container-fluid px-0" data-ng-controller="PrivateDatabaseConfigurationsCtrl as configurationsCtrl">
 
+    <div data-ovh-alert="{{alerts.main}}"></div>
+
     <div class="text-center" data-ng-if="configurationsCtrl.loading">
```
tremblaymath: It should be a ui-kit class to be reused across all managers. An align attribute on oui-spinner could be a possibility, it's a recurrent requirement.

client/app/private-database/configuration/private-database-configuration.html
```diff
@@ -1,76 +1,67 @@
...
-                                        data-ng-model="configuration.value"
-                                        data-ng-options="conf for conf in configuration.availableValues">
-                                </select>
-                                <span class="oui-icon oui-icon-chevron-down" aria-hidden="true"></span>
-                            </div>
-                        </div>
-                    </div>
-                </form>
-            </div>
+                <oui-button
```
tremblaymath: Should be replaced with oui-form-actions
Missing type=submit.
submit action should be on the form(ng-submit), not on submit button.
Submit button should be always available, never disabled.

client/app/private-database/configuration/private-database-configuration.html
```diff
@@ -1,76 +1,67 @@
...
-            <div data-ovh-alert="{{alerts.main}}"></div>
+    <div data-ng-if="!configurationsCtrl.loading">
+
+        <div class="alert alert-info" role="alert"
+             data-i18n-static="privateDatabase_configurations_general_postgresql_soon"
+             data-ng-if="configurationsCtrl.database.type === 'postgresql'">
+        </div>
+        <div class="alert alert-info" role="alert"
+             data-i18n-static="privateDatabase_capability_disabled_explanation"
+             data-ng-if="configurationsCtrl.database.type !== 'postgresql' && configurationsCtrl.configurations.length === 0">
```
tremblaymath: Use one time binding every time it's possible.
Several places in this file

client/app/private-database/configuration/private-database-configuration.html
```diff
@@ -1,76 +1,67 @@
...
             </div>
 
-            <div data-ng-if="configurationsCtrl.database.type !== 'postgresql'">
-                <h2 data-ng-bind-html="tr('privateDatabase_configurations_general_' + configurationsCtrl.database.type)"></h2>
+            <h2 data-ng-bind-html="tr('privateDatabase_configurations_general_' + configurationsCtrl.database.type)"></h2>
 
-                <div class="alert alert-warning" role="alert"
-                     data-i18n-static="privateDatabase_configuration_alert"
-                     data-ng-if="configurationsCtrl.edit.value">
+            <form name="db_configuration_form">
```
tremblaymath: form name should be in lower camel case

client/app/private-database/configuration/private-database-configuration.html
```diff
@@ -1,76 +1,67 @@
...
+                                    required
+                                    data-align="start">
+                            <span ng-bind="$item.text"></span>
+                        </oui-select>
+                    </oui-field>
+                    <oui-field label="{{configuration.key}}"
+                               size="s"
+                               ng-if="configuration.type === 'toggle'"
+                               help-text="{{configuration.description}}">
+                        <oui-radio-toggle-group model="configuration.selectedValue.id" required>
```
tremblaymath: Required isn't coherent, radio group should always have a selected value

client/app/private-database/configuration/private-database-configuration.html
JDownloader: I cannot use single binding there, because data is not loaded at controller instantiation

client/app/private-database/configuration/private-database-configuration.html
JDownloader: CX was ok with disabled if data is not modified, I will see we could use the ```oui-form-actions```

JDownloader: Yes, guidelines say it cannot be disabled

client/app/private-database/configuration/private-database-configuration.html
tremblaymath: ok! but there is other examples: `text="{{ tr('privateDatabase_configuration_confirm_btn') }}"`..

client/app/private-database/configuration/private-database-configuration.controller.js
```diff
@@ -24,41 +24,68 @@ angular.module("App").controller(
...
                 .getConfigurationDetails(this.productId)
                 .then((config) => {
-                    this.configurations = config.details;
-                    this.baseConfiguration = angular.copy(config.details);
+                    this.configurations = config.details.map((field) => this.convertAvailableValues(field));
                 })
-                .catch((err) => {
-                    this.alerter.alertFromSWS(this.$scope.tr("privateDatabase_configuration_error"), err, this.$scope.alerts);
+                .catch(() => {
+                    this.alerter.error(this.translator.tr("privateDatabase_configuration_error"), this.$scope.alerts.main);
```
tremblaymath: Logging error with $log.error(.. could be interesting for debugging and sentry

FredericEspiau: Would also be a good idea to actually retrieve the error from the API

client/app/private-database/configuration/private-database-configuration.controller.js
```diff
@@ -24,41 +24,68 @@ angular.module("App").controller(
...
+
+        static convertFieldAsToggle (field) {
+            field.type = "toggle";
+            field.selectedValue = { id: field.value };
+        }
+
+        getFieldDescriptionTranslated (field) {
+            const translationId = `privateDatabase_configuration_field_${field.key}`;
+            let description = this.translator.tr(translationId);
+            if (description.includes(translationId)) {
```
FredericEspiau: Not supported by IE11 : https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/String/includes

client/app/private-database/configuration/private-database-configuration.html
FredericEspiau: Having `text-center` as a ui-kit class is a nice idea but an align attribute would be against W3C and WHATWG specs

client/app/private-database/configuration/private-database-configuration.html
```diff
@@ -1,76 +1,60 @@
...
 
-            <div data-ng-if="configurationsCtrl.database.type !== 'postgresql'">
-                <h2 data-ng-bind-html="tr('privateDatabase_configurations_general_' + configurationsCtrl.database.type)"></h2>
+        <div data-ng-if="configurationsCtrl.configurations.length >= 1">
 
-                <div class="alert alert-warning" role="alert"
-                     data-i18n-static="privateDatabase_configuration_alert"
-                     data-ng-if="configurationsCtrl.edit.value">
-                </div>
+            <oui-message type="warning">
```
FredericEspiau: `data-type`

client/app/private-database/configuration/private-database-configuration.html
```diff
@@ -1,76 +1,60 @@
...
-            </button>
-            <button class="btn btn-block btn-default" type="button"
-                    data-i18n-static="privateDatabase_configuration_cancel_btn"
-                    data-ng-click="configurationsCtrl.configurations = configurationsCtrl.baseConfiguration; configurationsCtrl.edit.value = false"
-                    data-ng-disabled="database.expired"
-                    data-ng-if="configurationsCtrl.edit.value">
-            </button>
+            <form name="dbConfigurationForm" data-ng-submit="configurationsCtrl.updateConfigurations()">
+                <div data-ng-repeat="configuration in configurationsCtrl.configurations track by $index">
+                    <oui-field label="{{configuration.key}}"
```
FredericEspiau: `data-`

client/app/resources/i18n/privateDatabase/Messages_fr_FR.xml
```diff
@@ -101,6 +101,16 @@
    <translation id="privateDatabase_configuration_interactive_timeout" qtlid="370200">Le nombre de secondes que le serveur attend avant de fermer la connexion lorsqu'il n'y a pas d'activité lors d'une connexion en invite de commande.</translation>
    <translation id="privateDatabase_configuration_wait_timeout" qtlid="370213">Le nombre de secondes que le serveur attend avant de fermer la connexion lorsqu'il n'y a pas d'activité lors d'une connexion externe.</translation>
 
+   <translation id="privateDatabase_configuration_field_autocommit">Définit si les requêtes sont automatiquement commitées ou non.</translation>
```
FredericEspiau: Maybe you could use `soumettre` instead of `commiter`, if this is the same meaning

client/app/private-database/configuration/private-database-configuration.html
antleblanc: @FredericEspiau http://master.ui-kit.ovh/#!/oui-angular/decisions ;)

client/app/resources/i18n/privateDatabase/Messages_fr_FR.xml
```diff
@@ -101,6 +101,16 @@
...
+   <translation id="privateDatabase_configuration_field_event_scheduler">Permet de déclencher l’exécution de requêtes programmées directement dans le serveur MySQL.</translation>
+   <translation id="privateDatabase_configuration_field_innodb_buffer_pool_size">Choix de la taille de la mémoire tampon (en megabytes).</translation>
```
FredericEspiau: `mégaoctets`

client/app/private-database/configuration/private-database-configuration.html
tremblaymath: Ui-kit decisions are only for ui-kit repos.. for now, we keep data- for all managers

client/app/private-database/configuration/private-database-configuration.html
```diff
@@ -1,76 +1,60 @@
...
+                               ng-if="configuration.type === 'toggle'"
+                               help-text="{{configuration.description}}">
+                        <oui-radio-toggle-group model="configuration.selectedValue.id">
+                           <oui-radio ng-repeat="item in configuration.availableValues"
+                                      text="{{item.text}}"
+                                      value="item.id"></oui-radio>
+                        </oui-radio-toggle-group>
+                    </oui-field>
+                </div>
+                <oui-form-actions submit-text="{{ tr('privateDatabase_configuration_confirm_btn') }}"></oui-form-actions>
```
tremblaymath: Use one-time binding for translations and everywhere it's possible
data- for submit-text

client/app/private-database/configuration/private-database-configuration.html
antleblanc: ok @tremblaymath. thx.

client/app/private-database/configuration/private-database-configuration.html
tremblaymath: I'm curious, why it's against W3C and WHATWG?

FredericEspiau: According to @antleblanc I'm wrong.

I just assumed defining alignment outside of CSS would be viewed as wrong, as HTML's align attribute (https://www.w3.org/TR/html401/present/graphics.html#h-15.1.2) is deprecated, but according to him we can do whatever we want with components without having to worry with standards, I really don't know :smiley:

tremblaymath: I agree "defining alignment outside of CSS would be viewed as wrong"
IMHO "text-center" class has the same problem..
I could be better a more semantic class name e.g.: oui-page__content-loader without alignment concern
We will keep text-center as a temporary solution..

client/app/private-database/configuration/private-database-configuration.controller.js
JDownloader: I think ```$q.reject(...)``` is the prefered way of pushing error to Sentry in the web manager. That was not the point of this PR, but of course I can do that.

client/app/resources/i18n/privateDatabase/Messages_fr_FR.xml
JDownloader: This is a page for the experts, may be some french canadiens would like ```soumettre``` better but ```commiter``` is more meaningful. I think thine is fine.

client/app/private-database/configuration/private-database-configuration.html
FredericEspiau: No, the rules are to add `data-` to each attribute (cf https://github.com/ovh-ux/ovh-manager-web/pull/371#discussion_r177416164)

client/app/private-database/configuration/private-database-configuration.controller.js
```diff
@@ -24,41 +24,68 @@ angular.module("App").controller(
...
                 })
                 .finally(() => {
                     this.loading = false;
                 });
         }
 
+        convertAvailableValues (field) {
+            field.description = this.getFieldDescriptionTranslated(field);
+            field.availableValues = field.availableValues.map((value) => ({ id: value, text: value }));
+            if (field.key === "autocommit") {
```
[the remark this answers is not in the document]
FredericEspiau: At least if it's not null or undefined

client/app/private-database/configuration/private-database-configuration.controller.js
```diff
@@ -24,41 +24,68 @@ angular.module("App").controller(
...
+                description = field.description;
+            }
+            return description;
+        }
+
         updateConfigurations () {
             this.loading = true;
             this.edit.value = false;
-            const parameters = _.map(this.configurations, (conf) => ({ key: conf.key, value: conf.value }));
+            const parameters = this.configurations.map((conf) => ({ key: conf.key, value: conf.selectedValue.id }));
```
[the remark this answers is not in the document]
JDownloader: I try to escape that trend of [defensive software](https://en.wikipedia.org/wiki/Defensive_programming), I did verify and configurations could be an empty list but not ```undefined```.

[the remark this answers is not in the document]
tremblaymath: I agree with @JDownloader, if it's not an array if must raise an exception anyway we can't just ignore it.. error will be in Sentry logs.

jleveugle:
Can you resolve the conflict?

After that, feel free to add the right label on this pr and re-open it please :)

PR #404, opened 2018-05-24
fix(navbar): fix label property for brand

AxelPeter:
Fix property for label accessibility
https://master.ui-kit.ovh/#!/oui-angular/navbar


AxelPeter:
I've done the changes from the master branch instead

PR #411, opened 2018-06-05
Feature/new navbar notification

JDownloader:


antleblanc:
Related to:
- ovh-ux/ovh-manager-dedicated#368.
- ovh-ux/ovh-manager-telecom#726.

client/app/components/user/session.service.js
```diff
@@ -438,14 +448,20 @@ class SessionService {
...
-            user: this.user.getUser()
-        })
-            .then(({ user }) => getBaseNavbar(user))
+        return this.$q
+            .all({
+                translate: this.loadTranslations(),
+                user: this.user.getUser(),
+                notifications: this.navbarNotificationService.getNavbarContent()
+            })
+            .then(({ user, notifications }) => getBaseNavbar(user, notifications))
```
FredericEspiau: Whats `translate` for then ? You need it to be resolved before `getBaseNavbar` but you don't need the result, right ?

client/app/components/navbar/navbar-notification.service.js
```diff
@@ -0,0 +1,109 @@
...
+                this.OvhApiNotificationAapi.post({ acknowledged: toAcknowledge.map((x) => x.id) }).$promise.then(() => {
+                    toAcknowledge.forEach((sublink) => {
+                        sublink.acknowledged = true;
+                    });
+                });
+            }
+        }
+    }
+
+    _setRefreshTime (sublinks) {
```
FredericEspiau: Maybe you would refresh the elapsed time only when you open the notification center ? No need to compute time elapsed if nobody looks at it

FredericEspiau: Also, shouldn't it be done by the component itself ? The implementation would be the same, disregard the manager you're in

client/app/components/navbar/navbar-notification.service.js
```diff
@@ -0,0 +1,109 @@
...
+        } else if (!toUpdate.isActive && !toUpdate.updating && !linkClicked) {
+            toUpdate.updating = true;
+            this.OvhApiNotificationAapi.post({ acknowledged: [toUpdate.id] }).$promise.then(() => {
+                toUpdate.isActive = true;
+                toUpdate.acknowledged = true;
+            }).finally(() => { toUpdate.updating = false; });
+        }
+    }
+
+    convertSubLink (notification) {
```
FredericEspiau: `Object.assign` but I'm not sure Babel for the Web Manager has a polyfill

client/app/components/navbar/navbar-notification.service.js
```diff
@@ -0,0 +1,109 @@
...
+            id: "0",
+            date: moment(),
+            time: this._formatTime(moment()),
+            level: "error",
+            subject: this.translator.tr("common_navbar_notification_error_subject"),
+            description: this.translator.tr("common_navbar_notification_error")
+        }]);
+    }
+
+    /* eslint-disable class-methods-use-this */
```
marie-j: Maybe create a static method instead ?

client/app/components/navbar/navbar-notification.service.js
JDownloader: I don't want to add moment dependancy to ovh-ui-angular

client/app/components/user/session.service.js
JDownloader: probably, but it avoid problems because of all those single binding in the different managers and I want my code to be the same between all managers

client/app/components/navbar/navbar-notification.service.js
```diff
@@ -0,0 +1,109 @@
...
+    }
+    /* eslint-enable class-methods-use-this */
+
+    _toggleSublinkAction (toUpdate, linkClicked) {
+        if (toUpdate.isActive && !toUpdate.updating) {
+            toUpdate.updating = true;
+            this.OvhApiNotificationAapi.post({ completed: [toUpdate.id] }).$promise.then(() => {
+                toUpdate.isActive = false;
+                toUpdate.acknowledged = true;
+            }).finally(() => { toUpdate.updating = false; });
```
marie-j: No catching ?

client/app/components/navbar/navbar-notification.service.js
JDownloader: We don't have polyfill for the managers (dedicated inherit some polyfill from a non-ovh dependency, so it don't count). Of curse I could use lodash to redifine my Object, but there's no guideline about it and the result would be exactly the same.

client/app/components/navbar/navbar-notification.service.js
```diff
@@ -0,0 +1,109 @@
...
+        }).$promise);
+    }
+
+    getSubLinks () {
+        return this.getMessages().then((messages) => messages.map((message) => this.convertSubLink(message))).catch(() => [{
+            id: "0",
+            date: moment(),
+            time: this._formatTime(moment()),
+            level: "error",
+            subject: this.translator.tr("common_navbar_notification_error_subject"),
```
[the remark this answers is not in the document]
JDownloader: ´$translate.instant("KEY")´­ don't work in this manager

client/app/components/navbar/navbar-notification.service.js
JDownloader: because I cannot start a static method with underscore (eslint rule). I javascript I found that static functions are best when called outside of the main class like `myObject.staticFunction()`, but inside you would have to call `MyObject.staticFunction()`

client/app/components/navbar/navbar-notification.service.js
JDownloader: The feedback I give to the client is good, if the notification can't be updated it's not shown as updated. I don't think mark a notif as read is critical and need to display a real feedback when it fails.

PR #418, opened 2018-06-10
fix: chart now display correctly

FredericEspiau:
### Description of the Change

Before : charts scales were nonsensical, tooltips didn't work properly, there were no units, a useless zoom was present

Now: it's fixed !

SCFRCA-1109

### DIFF
--- a/client/app/private-database/metrics/private-database-metrics.controller.js
+++ b/client/app/private-database/metrics/private-database-metrics.controller.js
@@ -1,99 +1,78 @@
-angular.module("App").controller(
-    "PrivateDatabaseMetricsCtrl",
-    class PrivateDatabaseMetricsCtrl {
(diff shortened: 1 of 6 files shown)


client/app/private-database/metrics/private-database-metrics.controller.js
```diff
@@ -1,99 +1,78 @@
...
-            this.loaders = {
-                metrics: false
-            };
-
-            this.chart = {
-                memory: null,
-                connections: null,
-                time: null
-            };
+            this.charts = { };
```
antleblanc: `{ }` => `{}`

FredericEspiau: But I like it this way 😢

PR #419, opened 2018-06-11
fix(cloudweb): display runtimes in external multisite creation

cbourgois:
### Description of the Change

Allow the user to select the runtime in external multisite creation (cloudweb)

### DIFF
--- a/client/app/hosting/multisite/add/external/hosting-multisite-add-external.html
+++ b/client/app/hosting/multisite/add/external/hosting-multisite-add-external.html
@@ -57,6 +57,24 @@
         </div>
     </div>
 
+    <div class="form-group" data-ng-if="hosting.isCloudWeb">
+        <label class="control-label col-md-4" for="selectedRuntime" data-i18n-static="hosting_tab_DOMAINS_configuration_modify_step1_runtime"></label>
+        <div data-ng-if="loaders.runtimes">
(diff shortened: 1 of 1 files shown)


client/app/hosting/multisite/add/external/hosting-multisite-add-external.html
```diff
@@ -57,6 +57,26 @@
         </div>
     </div>
 
+    <div class="form-group" data-ng-if="hosting.isCloudWeb">
+        <label class="control-label col-md-4" for="selectedRuntime" data-i18n-static="hosting_tab_DOMAINS_configuration_modify_step1_runtime"></label>
+        <div data-ng-if="loaders.runtimes">
+            <oui-spinner data-size="s"></oui-spinner>
+        </div>
+
+        <div data-ng-if="!loaders.runtimes">
```
antleblanc: You can apply `data-ng-if="!loaders.runtimes"` directly to `<div class="col-md-8">`.

client/app/hosting/multisite/add/external/hosting-multisite-add-external.html
```diff
@@ -57,6 +57,26 @@
...
+            <div class="col-md-8">
+                <div class="oui-select oui-select_inline mb-0">
```
marie-j: Why not use `<oui-select>` component ?

PR #435, opened 2018-06-19
fix: fix guide tile

neolitec:
SCFRCA-1128 - Guides and Tips button display badly when opened

### DIFF
--- a/client/app/components/guides/guides.html
+++ b/client/app/components/guides/guides.html
@@ -1,22 +1,34 @@
-<div class="oui-tile mb-5" data-ng-if="!!guideConfiguration && guideConfiguration.length > 0">
-    <h3 class="oui-tile__title" data-ng-bind="guidesTitle"></h3>
+<div class="oui-tile oui-tile_collapsible mb-5"
+     data-ng-if="!!guideConfiguration && guideConfiguration.length > 0">
+    <h3 class="oui-tile__title"
+        data-ng-bind="guidesTitle">
+    </h3>
     <div class="oui-tile__body">
-        <div class="oui-tile__item" data-ng-repeat="guideSection in guideConfiguration track by $index">
-            <div class="oui-tile__item_button">
-                <button type="button" class="oui-tile__button oui-button oui-button_link oui-button_full-width d-flex text-wrap" data-ng-click="toggleGuides($index)"
-                        data-ng-class="{ 'showGuidesStatus': item.active }">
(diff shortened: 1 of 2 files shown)


client/app/components/guides/guides.less
```diff
@@ -1,13 +1,7 @@
...
-
-.lightbulb-guide {
-    button {
-        background-color: transparent;
-        color: @textColor;
-        padding: .5rem 0;
-        text-align: left;
-        white-space: normal;
-        width: 100%;
+.oui-tile {
```
antleblanc: Does it will be backported to OVH UI Kit? If yes, could you please adding a comment such as "TODO: should be removed when it will be available on the OVH UI Kit"

client/app/components/guides/guides.html
```diff
@@ -1,15 +1,13 @@
...
-                <button type="button" class="oui-tile__button oui-button oui-button_link oui-button_full-width d-flex text-wrap" data-ng-click="toggleGuides($index)"
-                        data-ng-class="{ 'showGuidesStatus': item.active }">
-                    <span data-ng-bind="tr(guideSection.title)"></span>
-                    <span data-ng-show="showGuidesStatus[$index]" class="fa fa-chevron-up my-auto ml-auto"></span>
-                    <span data-ng-show="!showGuidesStatus[$index]" class="fa fa-chevron-down my-auto ml-auto"></span>
-                </button>
-            </div>
+        <div class="oui-tile__item oui-tile__item_button" data-ng-repeat="guideSection in guideConfiguration track by $index">
+            <button type="button" class="oui-tile__button oui-button oui-button_link oui-button_full-width d-flex text-wrap" data-ng-click="toggleGuides($index)"
+                    data-ng-class="{ 'showGuidesStatus': item.active }">
```
antleblanc: Class `showGuidesStatus` is no longer used, so I think you can get rid of `data-ng-class`.

client/app/components/guides/guides.less
FredericEspiau: Pretty sure that most "TODO" are just comments that will be forgotten in near future

antleblanc: That's your opinion but a "TODO" comment is always appreciable if in a near future someone goes through here AND/OR decide to tackle some TODOs.

antleblanc:
@neolitec I finally updated your changes/commit by removing unused class `showGuidesStatus` and update commit message.

PR #467, opened 2018-07-13
fix(hosting): update password conditions in hosting module creation

cbourgois:
## Update password conditions in hosting module creation

MFRWW-626

### Description of the Change

Change password max length from `12`/`31` to `31`.

### DIFF
--- a/client/app/hosting/module/add/hosting-module-add.controller.js
+++ b/client/app/hosting/module/add/hosting-module-add.controller.js
@@ -2,7 +2,7 @@ angular
   .module('App')
   .controller(
     'HostingModuleCreateCtrl',
-    ($scope, $q, $stateParams, HostingModule, Alerter, Hosting, constants) => {
+    ($scope, $q, $stateParams, Alerter, constants, Hosting, HostingModule, HOSTING_MODULE) => {
       $scope.model = {
         databases: null,
         moduleTemplates: null,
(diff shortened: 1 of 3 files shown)


client/app/hosting/module/add/hosting-module-add.controller.js
```diff
@@ -19,6 +19,8 @@ angular
         path: '',
         minPasswdLength: 8,
         maxPasswdLength: 30,
+        minDBPasswordLength: 8,
```
antleblanc: Putting this in a `.constants.js` file make more sense.

client/app/hosting/module/add/hosting-module-add.html
```diff
@@ -182,12 +182,12 @@
                                data-i18n-static="hosting_configuration_tab_modules_create_step2_db_passwd"></label>
                         <div class="col-md-6">
                             <input type="password" autocomplete="off" class="form-control" id="dbPasswd" name="dbPasswd" required
-                                   data-ng-minlength="8"
-                                   data-ng-maxlength="model.databaseSelected.name ? 12 : 31"
+                                   data-ng-minlength="model.minDBPasswordLength"
```
antleblanc: `Database` instead of `DB` is more readable.

client/app/hosting/module/add/hosting-module-add.controller.js
cbourgois: OK, it's just a little fix, but I will put 
```
minPasswdLength: 8,
maxPasswdLength: 30,
minDBPasswordLength: 8,
maxDBPasswordLength: 31,
```
in a constants file.

client/app/hosting/module/add/hosting-module-add.html
cbourgois: We have both in this file `formModuleDB.dbPasswd` and `databaseHostedSelected`...
But ok I will change it.

PR #511, opened 2018-08-28
fix(emails): email details list

ganeshkumar1989:
MFRWW-829

### Requirements

The emails table is not getting populated, even if the data is being correctly fetched by the APIs.

### Description of the Change

The email data is being fetched correctly. But the de-structuring assignment was not being done properly, resulting in an error in the transform function. This has been fixed.

### DIFF
--- a/client/app/email-domain/delegate/email-domain-delegate.controller.js
+++ b/client/app/email-domain/delegate/email-domain-delegate.controller.js
@@ -120,16 +120,16 @@ angular.module('App').controller(
           email: this.Emails.getDelegatedEmail(item),
           usage: this.Emails.getEmailDelegatedUsage(item),
         })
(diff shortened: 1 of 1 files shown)


client/app/email-domain/delegate/email-domain-delegate.controller.js
```diff
@@ -120,16 +120,16 @@ angular.module('App').controller(
           email: this.Emails.getDelegatedEmail(item),
           usage: this.Emails.getEmailDelegatedUsage(item),
         })
-        .then(({ originalEmail, usage }) => {
-          const email = _(originalEmail).clone();
+        .then(({ email, usage }) => {
```
FredericEspiau: God I'm an idiot

PR #522, opened 2018-09-04
refactor(domain.zone): use oui-action-menu

TedKac:


client/app/domain/zone/ZONE.html
```diff
@@ -200,15 +200,13 @@
...
-                                        <button class="btn btn-icon" type="button"
-                                                data-ng-if="!(record.fieldType === 'NS' && !record.subDomain) && ctrlDomainTabZoneDns.allowModification"
-                                                data-linkedpopover="domain/zone/domain-zone-dns.popover.html"
-                                                data-linkedpopover-placement="left"
-                                                data-linkedpopover-remote="true"
-                                                data-linkedpopover-single="true"
-                                                data-linkedpopover-title="{{i18n.common_actions}}">
-                                            <span class="oui-icon oui-icon-gear_line oui-icon_small" aria-hidden="true"></span>
-                                        </button>
+                                        <oui-action-menu data-compact data-align="end"
```
antleblanc: In that way we agree that a cell can be empty which is not the best option. Isn't it? 

A better solution would be  to apply the `ngIf`directive on the parent `<td>` and increase a `colspan` attribute to one of siblings elements.

But as it was already implemented in that way, let's move on:)

FYI: next enhancement could be to rewrite this view by using the `<oui-datagrid>` component.

Anyway, thanks a lot!

PR #525, opened 2018-09-06
refactor(domain.dynhost.login): use oui-back-button

TedKac:


client/app/domain/dynhost/login/domain-dynhost-login.html
```diff
@@ -8,11 +8,9 @@
 
         <div data-ng-if="!ctrlDomainDynHostLogin.loading.init && !ctrlDomainDynHostLogin.hasResult">
             <div>
```
FredericEspiau: Could loose the `div`

PR #526, opened 2018-09-09
fix(domain.general.informations): remove alert message from domain without hosting

RenaudLo:
Hello

### Requirements : 

When you select a domain name in general informations. If the domain name hasn't web hosting associated, you have alert message : 

> There was an error when retrieving associated web hosting plans (This service does not exist)

The domain name hasn't web hosting so, it isn't necessary to show alert message but we can show all web hostings with the associated domain

### Description of the Change:

I create new function getHostingWithAttachedDomain based on api:/hosting/web/attachedDomain instead of getAttachedDomains because function getAttachedDomains return 404 if the service doesn't exist. 

I use new function getHostingWithAttachedDomain in domain-general-informations.controller.js

### Benefits

Help clients

### DIFF
--- a/client/app/domain/general-informations/domain-general-informations.controller.js
+++ b/client/app/domain/general-informations/domain-general-informations.controller.js
@@ -205,7 +205,7 @@ angular.module('App').controller(
     getAssociatedHosting(serviceName) {
       this.loading.associatedHosting = true;
       this.hostingAssociated = [];
-      return this.HostingDomain.getAttachedDomains(serviceName)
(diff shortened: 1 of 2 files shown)


euhmeuh:
We're sending this to Quality Check right now :)
Thank you for the contribution!

euhmeuh:
Ref: MFRWW-1127

RenaudLo:
Hello,

The bug is fix with PR #537.

Thanks!

I close the PR.

Can you close MANAGER-752?

antleblanc:
MANAGER-752 is now closed! Thanks!

PR #560, opened 2018-09-25
refactor(private.databse.state): use oui-status & class

TedKac:
## refactor(private.databse.state): use oui-status & class


antleblanc:
Thanks @TedKac for those enhancements.

I strongly encourage you to have a look at the [`<oui-tile>`](http://master.ui-kit.ovh/#!/oui-angular/tile) component.

PR #565, opened 2018-09-27
style(hosting.general.informations): move progress bar in tile

TedKac:

### DIFF
--- a/client/app/hosting/general-informations/GENERAL_INFORMATIONS.html
+++ b/client/app/hosting/general-informations/GENERAL_INFORMATIONS.html
@@ -84,6 +84,58 @@
                                         </button>
                                     </oui-action-menu>
                                 </li>
+                                <li class="oui-tile__item">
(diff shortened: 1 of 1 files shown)


client/app/hosting/general-informations/GENERAL_INFORMATIONS.html
```diff
@@ -84,6 +84,58 @@ <h3 class="oui-header_4 oui-tile__title" data-translate="hosting_dashboard_gener
                                         </button>
                                     </oui-action-menu>
                                 </li>
+                                <li class="oui-tile__item">
+                                    <div data-ng-if="!hosting.isExpired"
+                                         data-ng-show="hosting.quotaPercent != null && hosting.quotaPercent >= 0">
```
JeremyDec: `hosting.quotaPercent` can be null or undefined ? Or it can be null only ? If so please use strict equality (`hosting.quotaPercent !== null`)

PR #668, opened 2018-10-18
fix(privatedb.order): display warning for 512MB only

euhmeuh:
MANAGER-1766

### DIFF
--- a/client/app/private-database/order/sql-database-order-controller.js
+++ b/client/app/private-database/order/sql-database-order-controller.js
@@ -203,6 +203,12 @@ angular.module('App').controller(
       }
     }
 
+    shouldDisplayFreeDbWarning() {
+      return this.isPrivateDb()
+        && _.get(this.model, 'ram') === '512'
+        && _.get(this.selectedHosting, 'stillHasFreeDbOffer');
+    }
+
     /*
        * DURATION
        */
--- a/client/app/private-database/order/sql-database-order.html
+++ b/client/app/private-database/order/sql-database-order.html
@@ -140,7 +140,7 @@
                     </div>
 
                     <oui-message type="warning"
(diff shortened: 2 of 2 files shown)


client/app/private-database/order/sql-database-order.html
```diff
@@ -140,7 +140,7 @@ <h2 class="mt-5" data-translate="privateDatabase_order_sql_customize_title"></h2
                     </div>
 
                     <oui-message type="warning"
-                                 data-ng-if="$ctrl.isPrivateDb() && $ctrl.selectedHosting && $ctrl.selectedHosting.stillHasFreeDbOffer">
+                                 data-ng-if="$ctrl.shouldDisplayFreeDbWarning()">
```
FredericEspiau: `data-` right above 😄

client/app/private-database/order/sql-database-order-controller.js
```diff
@@ -203,6 +203,12 @@ angular.module('App').controller(
       }
     }
 
+    shouldDisplayFreeDbWarning() {
+      return this.isPrivateDb()
+        && _.get(this.model, 'ram') === '512'
```
FredericEspiau: `this.model.ram` and `this.selectedHosting. stillHasFreeDbOffer ` can be empty at this point ?

client/app/private-database/order/sql-database-order.html
euhmeuh: Bien vu x)

client/app/private-database/order/sql-database-order-controller.js
euhmeuh: The method is called from the DOM in the digest cycle, so it can be called before any promise.

PR #675, opened 2018-10-18
fix(email.rules): handle rules with special names

FredericEspiau:
## Handle rule with special characters

MANAGER-1771

Also fixed the "Other" dropdown issue when clicking on the cross button
Also refactored all the code for rules & delegated rules

### DIFF
--- a/client/app/email-domain/email-domain.service.js
+++ b/client/app/email-domain/email-domain.service.js
@@ -303,22 +303,16 @@ angular.module('services').service(
      * @param rules
      */
     createFilter(serviceName, accountName, filter, rules) {
-      const defer = this.$q.defer();
-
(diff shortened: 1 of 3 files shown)


client/app/email-domain/email-domain.service.js
```diff
@@ -487,20 +458,11 @@ angular.module('services').service(
...
-        .all(promises)
-        .then(() => defer.resolve())
-        .catch(data => defer.resolve(data));
-
-      return defer.promise;
+      return this.$q
+        .all(
+          rulesId.map(id => this.deleteRule(serviceName, accountName, filterName, id)),
+        )
+        .catch(data => data);
```
marie-j: Isn't that the default behavior ?

PR #706, opened 2018-10-30
refactor(components.converter): move to webUniverseComponents

cbourgois:
### Description of the Change

e27fde06 - refactor(components.converter): remove unused function
b4775a45 - refactor(components.converter): move to webUniverseComponents

ref: OM-153

/cc @jleveugle @frenautvh @antleblanc


client/app/private-database/order/sql-database-order-controller.js
```diff
@@ -14,7 +14,7 @@ angular.module('App').controller(
       this.$window = $window;
 
       this.atInternet = atInternet;
-      this.converterService = ConverterService;
```
antleblanc: Shouldn't be also `this.WucConverterService`? According to this similarity changehttps://github.com/ovh-ux/ovh-manager-web/pull/706/files#diff-fd7cb8f42f3c72d2a0dad6185fc41500L20.

cbourgois: Should be great :)

PR #736, opened 2018-11-07
Fix/tooltip hosting limit decimal

varun257:
### Description of the Change
fix number to show 2 decimal places
fix translation error

### Benefits

Looks better

### Possible Drawbacks
None

### Applicable Issues

MBE-15


client/app/hosting/general-informations/GENERAL_INFORMATIONS.html
```diff
@@ -79,8 +79,8 @@ <h3 class="oui-header_4 oui-tile__title" data-translate="hosting_dashboard_gener
                                                 data-ng-click="setAction('offer/upgrade/hosting-offer-upgrade')"
                                                 data-track-on="click"
                                                 data-track-name="web::hostname::general-informations::change-offer"
-                                                data-track-click="action"
-                                                data-translate="hosting_dashboard_service_change_offer">
+                                                data-track-click="action">
+                                                {{'hosting_dashboard_service_change_offer' | translate}}
```
antleblanc: If it's the `translate` directive that doesn't work properly with `track-*` directives, I encourage you to prefer create a `<span>` element instead of `{{`.

```suggestion
                                                <span data-translate="hosting_dashboard_service_change_offer"></span>
```

PR #790, opened 2018-11-23
fix(displayname-fix): fix display names of email and domain

varun257:
### Description of the Change

Add API to fetch displayname in email
Show displayname if available else show name

### Benefits

Display name is consistent with left navbar

### Possible Drawbacks

None

### Applicable Issues

resolves MBP-137, MBP-138

### DIFF
--- a/client/app/dns-zone/dns-zone.controller.js
+++ b/client/app/dns-zone/dns-zone.controller.js
@@ -85,7 +85,7 @@ angular.module('App').controller(
           this.domain = product;
 
           return this.Domain.getZoneByZoneName(this.domain.name).then((zoneInfos) => {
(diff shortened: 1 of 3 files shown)


client/app/dns-zone/dns-zone.controller.js
```diff
@@ -85,7 +85,7 @@ angular.module('App').controller(
           this.domain = product;
 
           return this.Domain.getZoneByZoneName(this.domain.name).then((zoneInfos) => {
-            this.domain.displayName = this.domain.name;
+            this.domain.displayName = this.domain.displayName || this.domain.name;
```
FredericEspiau: this doesn't have to be in the `then` then :)

PR #807, opened 2018-11-30
fix(hosting): avoid further api calls if hosting cannot be found

marie-j:
### Description of the Change

If hosting was not found, other calls were made that returns the same error (404 Not found). Prevent those calls from being made in that case. 

Related to MBP-145

### DIFF
--- a/client/app/hosting/general-informations/GENERAL_INFORMATIONS.controller.js
+++ b/client/app/hosting/general-informations/GENERAL_INFORMATIONS.controller.js
@@ -2,22 +2,28 @@ angular.module('App').controller(
   'hostingGeneralInformationsCtrl',
   class HostingGeneralInformationsCtrl {
     constructor(
+      $q,
       $scope,
       $stateParams,
       $translate,
       Alerter,
-      hostingSSLCertificate,
+      Hosting,
       HostingLocalSeo,
       HostingRuntimes,
+      hostingSSLCertificate,
+      OvhApiScreenshot,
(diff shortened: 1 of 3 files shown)


client/app/hosting/general-informations/GENERAL_INFORMATIONS.controller.js
```diff
@@ -35,7 +42,11 @@ angular.module('App').controller(
       };
 
       this.$scope.$on('hosting.ssl.reload', () => this.retrievingSSLCertificate());
-      return this.retrievingSSLCertificate()
+      return this.$q.all({
```
FredericEspiau:
```suggestion
      return this.$q.all([
          this.getUserLogsToken(),
           this.getScreenshot(),
      ])
```

client/app/hosting/hosting.controller.js
```diff
@@ -238,167 +222,32 @@ angular
...
-          )
-          .then((screenshot) => {
-            $scope.screenshot = screenshot;
-          })
-          .finally(() => {
-            $scope.loadingScreenshot = false;
-          });
+      function setUrchin() {
+        const [, cluster] = _.get($scope.hostingProxy, 'cluster', '').split('cluster');
+        if (cluster && parseInt(cluster, 10) >= 20) {
```
FredericEspiau: Watch out because `parseInt` acts weird sometimes: `parseInt('100a2', 10) === 100`

PR #809, opened 2018-12-03
fix(status-fix): fix status for cloud offers

varun257:
### Description of the Change

Add API to fetch hosting
Based on resource type discern status

### Benefits

Shows correct status for cloud and non cloud offers

### Possible Drawbacks

None

### Applicable Issues

resolves MBP-182

### DIFF
--- a/client/app/hosting/task/hosting-task.controller.js
+++ b/client/app/hosting/task/hosting-task.controller.js
@@ -37,13 +37,13 @@ angular.module('App').controller(
           offset - 1,
         ),
       })
-        .then((resp) => {
-          this.hosting = resp.hosting;
-          this.tasksList = resp.tasks;
+        .then(({ hosting, tasks }) => {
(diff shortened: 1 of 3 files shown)


client/app/hosting/task/hosting-task.controller.js
```diff
@@ -28,17 +29,21 @@ angular.module('App').controller(
...
-          this.tasksList = tasks;
+      return this.$q.all({
+        hosting: this.Hosting.getHosting(this.$stateParams.productId),
+        tasks: this.Hosting.getTasksList(
+          this.$stateParams.productId,
+          pageSize,
+          offset - 1,
+        ),
+      })
+        .then((resp) => {
```
antleblanc: You can use the destructuring parameters syntax:
```suggestion
        .then(({ hosting, tasks }) => {
```

PR #818, opened 2018-12-06
fix(hosting.databases): update upgrade text

ganeshkumar1989:
Close MBE-165

### Requirements

The specs given in hosting>databases upgrade message is wrong

Nos offres vous permettrons d'obtenir :

Un FTP d'au moins 100 Go
Un minimum de 10 adresses e-mail de 5 Go
Une base de données commençant à 200 Mo

This information does not take into account the offer "Web Kimsufi". The exact information should be : 

Nos offres vous permettrons d'obtenir :

Un FTP d'au moins 1 Go
Un minimum de 2 adresses e-mail de 5 Go
Une base de données commençant à 100 Mo

### Description of the Change

The above text change is done


FredericEspiau:
This text should be displayed only if the offer is 'Web Kimsufi'

ganeshkumar1989:
Kimsufi offer is always available and it is the the smallest Webhosting offer available. Hence it should always be shown. This is my understanding. Please correct me if I am wrong :)

FredericEspiau:
@ganeshkumar1989 I think this text should be displayed only if the offer is `Web Kimsufi` and the previous text should be displayed otherwise, but I'm not sure

ganeshkumar1989:
@FredericEspiau , I just confirmed with Teddy that Web Kimsufi is always available and is the smallest offer and that the text displayed should match the configuration of that of Kimsufi

PR #821, opened 2018-12-10
fix(tooltip-fix): disable tooltip for webhosting

varun257:
### Description of the Change

get resource type of the solution
disbale tooltip based on this

### Benefits

Helps hide tooltip for webhosting offers

### Possible Drawbacks

None

### Applicable Issues

Resolves MBP-24

### DIFF
--- a/client/app/hosting/database/DATABASE_LIST.html
+++ b/client/app/hosting/database/DATABASE_LIST.html
@@ -150,6 +150,7 @@
                             <td class="word-break"
                                 data-ng-bind="element.name"
                                 data-uib-tooltip="{{element.server}}"
+                                data-tooltip-enable="ctrlTabDatabases.resourceType !== 'cloud'"
                                 data-tooltip-append-to-body="true">
                             </td>
                             <td data-ng-if="!hosting.isCloudWeb">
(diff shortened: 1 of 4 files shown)


client/app/hosting/database/hosting-database.controller.js
```diff
@@ -187,6 +187,20 @@ angular.module('App').controller(
       this.bddTemplate = 'hosting/database/dump/DUMPS.html';
     }
 
+    getResourceType() {
+      return this.hostingService
+        .getHosting(this.$stateParams.productId)
+        .then((config) => {
```
antleblanc:
```suggestion
        .then(({ resourceType }) => {
```

client/app/hosting/database/DATABASE_LIST.html
```diff
@@ -150,6 +150,7 @@
                             <td class="word-break"
                                 data-ng-bind="element.name"
                                 data-uib-tooltip="{{element.server}}"
+                                data-tooltip-enable="ctrlTabDatabases.resourceType !== 'cloud'"
```
antleblanc: Does [`<oui-tooltip>`](https://ovh-ux.github.io/ovh-ui-kit-documentation/#!/oui-angular/tooltip) component would be suitable here?

varun257: I did try and use `<oui-tooltip>` but `enable/disable` is not provided by it. Cant be used unless we separate text and tooltip(may be with icon)

PR #877, opened 2019-01-16
fix(delete-button): delete subscription button

varun257:
### Description of the Change

change name of unsubscribe button

### Benefits

Helps make it clear for user that subscription will be removed

### Possible Drawbacks

None

### Applicable Issues

resolves MBP-228


cbourgois:
I think your commit should be something like : 
```
fix(hosting): rename delete subscription button label
```

antleblanc:
> I think your commit should be something like :
> 
> ```
> fix(hosting): rename delete subscription button label
> ```

@cbourgois  I will handle that update with the `squash and merge` feature;)

PR #885, opened 2019-01-21
fix(domains-zone-cnme): validation for cname records

ganeshkumar1989:
Close MBE-189

### Requirements

A CNAME record is not allowed to coexist with any other data. In other words, if suzy.podunk.xx is an alias for sue.podunk.xx, you can't also have an MX record for suzy.podunk.xx, or an A record, or even a TXT record.

### Description of the Change

The validation for records have been added so that they are not created by the same sub-domain as an existing CNAME record or vice-versa.


ganeshkumar1989:
Closing this PR since we are taking a different approach to handle this issue. New PR: https://github.com/ovh-ux/ovh-manager-web/pull/887

PR #886, opened 2019-01-22
fix(domain): domain tabs tracking

varun257:
### Description of the Change

Add tracking to header tabs of domain

### Benefits

Helps track events on tabs

### Possible Drawbacks

None

### Applicable Issues

resolves MBE-34

### DIFF
--- a/client/app/domain/domain-tabs.controller.js
+++ b/client/app/domain/domain-tabs.controller.js
@@ -1,12 +1,14 @@
 angular.module('App').controller(
   'DomainTabsCtrl',
   class DomainTabsCtrl {
-    constructor($scope, $q, $location, $stateParams, $translate, Alerter, Domain, DOMAIN, User) {
+    constructor($scope, $q, $location, $stateParams, $translate,
+      atInternet, Alerter, Domain, DOMAIN, User) {
       this.$scope = $scope;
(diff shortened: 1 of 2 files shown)


client/app/domain/domain-tabs.controller.js
```diff
@@ -1,12 +1,14 @@
 angular.module('App').controller(
   'DomainTabsCtrl',
   class DomainTabsCtrl {
-    constructor($scope, $q, $location, $stateParams, $translate, Alerter, Domain, DOMAIN, User) {
+    constructor($scope, $q, $location, $stateParams, $translate,
```
FredericEspiau: Could you please reorder the services ? :)

client/app/domain/domain-tabs.controller.js
```diff
@@ -131,6 +133,12 @@ angular.module('App').controller(
       } else {
         this.selectedTab = this.defaultTab;
       }
+
+      this.atInternet.trackClick({
```
FredericEspiau: Did you check if this work ? Because the expected names are very precise and I think the name of each tab is in UPPERCASE :)

varun257: yup all tabs are in uppercase, will verify this

PR #887, opened 2019-01-22
fix(domains-zone-cnme): validation for cname records

ganeshkumar1989:
Close MBE-189

### Requirements

A CNAME record is not allowed to coexist with any other data. In other words, if suzy.podunk.xx is an alias for sue.podunk.xx, you can't also have an MX record for suzy.podunk.xx, or an A record, or even a TXT record.

### Description of the Change

The validation for records have been added so that they are not created by the same sub-domain as an existing CNAME record or vice-versa.


client/app/domain/zone/record/resume.html
```diff
@@ -19,30 +19,23 @@
...
-            </ul>
+        <div data-ng-if="ctrl.model.fieldType === 'CNAME'"
+             data-translate="domain_configuration_dns_entry_add_step3_check_warning2">
+        </div>
+        <div data-ng-if="ctrl.model.fieldType !== 'CNAME'"
+             data-translate="domain_configuration_dns_entry_add_step3_check_warning3">
         </div>
     </div>
 
+    <div class="alert alert-warning mt-4" role="alert"
```
antleblanc: Could be turned later into a [`<oui-message>`](https://ovh-ux.github.io/ovh-ui-kit-documentation/#!/oui-angular/message) element:)

PR #892, opened 2019-01-28
fix(domain.zone): format public key on change

JeremyDec:
MANAGER-2119


client/app/domain/zone/record/domain-zone-record.controller.js
```diff
@@ -622,6 +625,10 @@ angular.module('App').controller(
       }
     }
 
+    formatPublicKey() {
+      this.model.target.p = this.model.target.p.replace(/\n/g, '');
```
marie-j: Could this be renamed to something more explicit than `p` ?

PR #956, opened 2019-03-08
Feat/mx plan

lizardK:
### Description of the Change
Add some features for new MX Plan products : 
- Mailing-lists
- Redirections
- Improve dashboard

### Related
- [x] ovh-ux/ovh-module-emailpro#191

### References 
MBP-261


client/app/components/user/session.service.js
```diff
@@ -146,6 +146,10 @@ class SessionService {
       products.emails,
       elt => angular.lowercase(elt.displayName || elt.name),
     );
+    const emailProMXPlanProducts = _.sortBy(
+      products.emailsProMXPlan,
+      elt => angular.lowercase(elt.displayName || elt.name),
```
antleblanc: Quick note: `angular.lowercase` has been removed on the 1.7 version

ref: https://docs.angularjs.org/guide/migration#migrating-from-1-6-to-1-7

PR #965, opened 2019-03-13
feat(web-multiste): regenerate ssl

varun257:
### Description of the Change

add polling
add spinner and disable button

### Benefits

Gives better feedback to user

### Possible Drawbacks

None

### Applicable Issues

resolves MBP-350


package.json
```diff
@@ -124,7 +124,7 @@
     "ovh-module-exchange": "^9.3.4",
     "ovh-module-office": "^7.0.7",
     "ovh-module-sharepoint": "^7.1.4",
-    "ovh-ui-angular": "^2.24.1",
+    "ovh-ui-angular": "2.25.1",
```
antleblanc: Could you please keep the `^` in order to be able to upgrade to minor releases?

package.json
```diff
@@ -124,7 +124,7 @@
     "ovh-module-exchange": "^9.3.4",
     "ovh-module-office": "^7.0.7",
     "ovh-module-sharepoint": "^7.1.4",
-    "ovh-ui-angular": "^2.24.1",
+    "ovh-ui-angular": "^2.25.1",
```
Jisay: To avoid bad surprise, can you bump also `ovh-ui-kit ` to `v2.25.1`?

varun257: Have bumped both to the latest releases which is 2.26.1

PR #966, opened 2019-03-14
feat(domain): replace wuc-ovh-tabs with oui-header-tabs

marie-j:
### Description of the Change

Before: Domains used `wuc-ovh-tabs`
Now: Domains uses oui-header-tabs and UI Router states 

MANAGER-2343


client/app/domain/dns/domain-dns.state.js
```diff
@@ -0,0 +1,14 @@
+const state = {
+  url: '/dns',
+  views: {
+    domainView: {
+      templateUrl: 'domain/dns/DNS.html',
```
frenautvh:
```javascript
import mytemplate from "foo/bar.html";
// ...
domainView: {
  template: mytemplate,
}
```
that way webpack will report errors if template is missing

cbourgois:
```javascript
import template from "foo/bar.html";
// ...
domainView: {
  template,
}
```

client/app/domain/domain.controller.js
```diff
@@ -192,6 +194,10 @@ angular.module('App').controller(
         .finally(() => {
           this.loading.domainInfos = false;
           this.$scope.$broadcast('domain.refreshData.done');
+
+          if (!this.domain.isExpired) {
+            this.$state.go(`${this.getState()}.information`, this.$stateParams);
```
JeremyDec: `return` ?

marie-j: As we're in the `finally` hook, the value will never be used (when chaining we'll get the result from the `then`/`catch` hook )

PR #987, opened 2019-04-05
Increase max dynhost password length from 12 to 31

trnsnt:
### Description of the Change

Increase max dynhost password length from 12 to 31 to be coherent with API.

### Benefits

Be coherent with API.

### Possible Drawbacks
None

### Applicable Issues
None


antleblanc:
Thanks @trnsnt for your contribution.

Do you know if theses information can be retrieve directly from the API instead?

It will help if theses data changes over the time.

Could you also please target `develop` branch?

Thanks!

trnsnt:
Hello @antleblanc, for now there is no way to get the password min / max length from the API. But it's something relatively stable over the time.
Moved to target develop !

Thx

PR #1050, opened 2019-05-02
fix(domain-dashboard): fix ssl order button

varun257:
### Description of the Change

enable add ssl button for all cases

### Possible Drawbacks

None

### Applicable Issues

resolves MBP-349

### DIFF
--- a/client/app/hosting/general-informations/GENERAL_INFORMATIONS.html
+++ b/client/app/hosting/general-informations/GENERAL_INFORMATIONS.html
@@ -199,8 +199,7 @@
                         <oui-action-menu
                             data-compact data-align="end"
                             data-ng-if="!$ctrl.isRetrievingSSLCertificate && ($ctrl.canOrderSSLCertificate() || $ctrl.canRegenerateSSLCertificate() || $ctrl.canDeleteSSLCertificate())">
-                            <oui-action-menu-item data-ng-if="$ctrl.canOrderSSLCertificate()"
-                                data-text="{{ ::'hosting_dashboard_service_order_ssl' | translate }}"
+                            <oui-action-menu-item data-text="{{ ::'hosting_dashboard_service_order_ssl' | translate }}"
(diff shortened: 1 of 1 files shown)


client/app/hosting/general-informations/GENERAL_INFORMATIONS.html
```diff
@@ -199,8 +199,7 @@
                         <oui-action-menu
                             data-compact data-align="end"
                             data-ng-if="!$ctrl.isRetrievingSSLCertificate && ($ctrl.canOrderSSLCertificate() || $ctrl.canRegenerateSSLCertificate() || $ctrl.canDeleteSSLCertificate())">
-                            <oui-action-menu-item data-ng-if="$ctrl.canOrderSSLCertificate()"
-                                data-text="{{ ::'hosting_dashboard_service_order_ssl' | translate }}"
+                            <oui-action-menu-item data-text="{{ ::'hosting_dashboard_service_order_ssl' | translate }}"
```
FredericEspiau: is `canOrderSSLCertificate ` still used ?

FredericEspiau: What's the point of `$ctrl.canOrderSSLCertificate()` if it doesn't determine when an user can order a certificate ? Are you sure you can order a certificate all the time ?

varun257: Yes Button should be available everytime was the requirement since user can order a new certificate before the current one expires and also covers a use case to upgrade existing certificate.

FredericEspiau: So shouldn't `canOrderSSLCertificate` be deleted ?

antleblanc: Thanks @varun257 for the update!

PR #1097, opened 2019-06-05
fix(components.navbar): specify the right state for domain and hosting

antleblanc:
# Specify the right state for domain and hosting

### :bug: Bug Fix

8097bbc - fix(components.navbar): specify the right state for domain and hosting

ref: DTRUN-1816

### DIFF
--- a/client/app/components/navbar/navbar.service.js
+++ b/client/app/components/navbar/navbar.service.js
@@ -36,7 +36,7 @@ export default class Navbar {
   }
 
   getSublinks(section, products) {
-    if (section.name === 'domain') {
+    if (section.name === 'domains') {
       return this.getDomainsMenu(products.domains);
     }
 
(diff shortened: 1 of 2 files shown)


client/app/components/navbar/navbar.service.js
```diff
@@ -81,15 +80,17 @@ export default class Navbar {
...
-      subLinks: _.map(domain.subProducts, subDomain => ({
-        name: subDomain.name,
-        title: subDomain.displayName,
-        state: 'app.domain.alldom',
-        stateParams: {
-          allDom: domain.displayName,
-          productId: subDomain.name,
-        },
-      })),
+      subLinks: domain.subProducts
```
FredericEspiau: Weird, I thought `Lodash`'s `map` was immune to this kind of errors

antleblanc: I will take a closer look to `<oui-navbar-menu>` because without doing that changes, I'm end up with an empty domains' list.

PR #1128, opened 2019-06-21
fix(hosting) allow cloudweb cdn perf as free option as in business model

VincentCasse:
### Description of the Change

CloudWeb offers include free cdn option.
Inside customer panel, we sell it.

MBP-374

### Possible Drawbacks

Can break the CDN option sell

### DIFF
--- a/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
+++ b/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
@@ -30,7 +30,7 @@ angular
           }
           $scope.loading.model = true;
           HostingOptionOrder.getOrderEnums('hosting.web.CdnOfferEnum').then((models) => {
-            if (!$scope.isPerfOffer()) {
+            if (!$scope.isPerfOffer() and !$scope.isCloudWebOffer() ) {
               $scope.availableOffers = models.filter(offer => offer !== 'CDN_BUSINESS_FREE');
             } else {
               $scope.availableOffers = ['CDN_BUSINESS_FREE'];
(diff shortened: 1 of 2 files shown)


client/app/hosting/cdn/order/hosting-cdn-order.controller.js
```diff
@@ -30,7 +30,7 @@ angular
           }
           $scope.loading.model = true;
           HostingOptionOrder.getOrderEnums('hosting.web.CdnOfferEnum').then((models) => {
-            if (!$scope.isPerfOffer()) {
+            if (!$scope.isPerfOffer() and !$scope.isCloudWebOffer() ) {
```
frenautvh: missing offer parameter

client/app/hosting/hosting.service.js
```diff
@@ -121,6 +121,15 @@
         return /^perf.+$/.test(offer);
       }
 
+      /**
+             * Is hosting cloudweb offer
```
marie-j: The indentation seems a bit off here

client/app/hosting/hosting.service.js
```diff
@@ -121,6 +121,15 @@
...
+             * @param {string} offer
+             * @returns {boolean}
+             */
+      static isCloudWebOffer(offer) {
+        return _(offer).includes('CLOUD');
```
marie-j: Prefer using `_.includes(offer, 'CLOUD')` instead

euhmeuh:
This should be enough: https://github.com/ovh-ux/ovh-manager-web/pull/1130

PR #1129, opened 2019-06-21
Translation missing

VincentCasse:
### Benefits

Better experience

MBP-382


client/app/hosting/cdn/order/hosting-cdn-order.controller.js
```diff
@@ -30,7 +30,7 @@ angular
           }
           $scope.loading.model = true;
           HostingOptionOrder.getOrderEnums('hosting.web.CdnOfferEnum').then((models) => {
-            if (!$scope.isPerfOffer()) {
+            if (!$scope.isPerfOffer() and !$scope.isCloudWebOffer() ) {
```
frenautvh: you're calling this function without offer argument ?

VincentCasse: You're right, one definition is missing in the local controller of order cdn part.

I just duplicate the behavior of isPerfOffer

VincentCasse: (and fixed)

VincentCasse: I also see my two branches was mixed up. For this part, please refer to my another PR (1128)

client/app/hosting/translations/Messages_fr_FR.json
```diff
@@ -270,6 +270,24 @@
...
+  "hosting_tab_CRON_table_language_RUBY2_4": "Ruby 2.4",
+  "hosting_tab_CRON_table_language_ruby2_5": "Ruby 2.5",
+  "hosting_tab_CRON_table_language_RUBY2_5": "Ruby 2.5",
+  "hosting_tab_CRON_table_language_ruby2_6": "Ruby 2.6",
+  "hosting_tab_CRON_table_language_RUBY2_6": "Ruby 2.6",
+  "hosting_tab_CRON_table_language_python2": "Python 2",
+  "hosting_tab_CRON_table_language_PYTHON2": "Python 2",
+  "hosting_tab_CRON_table_language_python3": "Python 3",
+  "hosting_tab_CRON_table_language_PYTHON3": "Python 3",
+  "hosting_tab_CRON_table_language_node8": "Node 8",
```
antleblanc: It shouldn't be this instead:

```suggestion
  "hosting_tab_CRON_table_language_node8": "Node.js 8",
```

?

marie-j: We should probably think of another way to display those instead of having to add translations each time, can be handled in another PR (Ticket reference created for this: MANAGER-2973)

VincentCasse: I'd just change Node by Node.js
For the improvment of translation system, i let the works for frontend teams ;)

PR #1164, opened 2019-07-10
fix(domain-options): addition of options tile

ganeshkumar1989:
MBP-460

### Requirements

* The APIs for domain options have been added. Hence the domain options have to be exposed to the user via a new tile in the domain dashboard page.
* Currently, only dnsAnycast is supported. This the user will only be able to delete the option, if it is activated.

### Description of the Change

The domain options tile has been added as per requirement.

### DIFF
--- a/client/app/domain/domain.service.js
+++ b/client/app/domain/domain.service.js
@@ -1177,6 +1177,40 @@ angular.module('services').service(
       });
     }
 
+    // --------------------- Options ----------------------------
+
+    /**
+     * Get all options
+     * @param {string} serviceName
+     */
+    getOptions(serviceName) {
+      return this.OvhHttp.get(`/domain/${serviceName}/option`, {
+        rootPath: 'apiv6',
+      });
+    }
+
+    /**
+     * Get the option details
+     * @param {string} serviceName
+     * @param {string} option
+     */
+    getOption(serviceName, option) {
+      return this.OvhHttp.get(`/domain/${serviceName}/option/${option}`, {
+        rootPath: 'apiv6',
+      });
+    }
+
+    /**
+     * Delete option
+     * @param {string} serviceName
+     * @param {string} option
+     */
+    deleteOption(serviceName, option) {
+      return this.OvhHttp.delete(`/domain/${serviceName}/option/${option}`, {
+        rootPath: 'apiv6',
+      });
+    }
+
     // --------------------- Glue registry ----------------------
 
     /**
--- a/client/app/domain/general-informations/GENERAL_INFORMATIONS.html
+++ b/client/app/domain/general-informations/GENERAL_INFORMATIONS.html
@@ -1,7 +1,7 @@
 <div class="container-fluid px-0" data-ng-controller="DomainTabGeneralInformationsCtrl as $ctrl">
     <div data-ovh-alert="{{alerts.main}}"></div>
     <div class="row d-lg-flex">
-        <div class="col-xm-12 col-md-4 mb-5">
+        <div class="col-md-4 mb-5">
             <oui-tile
                 data-heading="{{ ::'domain_dashboard_general_informations' | translate }}"
                 class="h-100"
@@ -130,7 +130,7 @@
                 </oui-tile-definition>
             </oui-tile>
         </div>
-        <div class="col-xm-12 col-md-4 mb-5">
+        <div class="col-md-4 mb-5">
             <oui-tile
                 data-heading="{{ ::'domain_dashboard_security' | translate }}"
                 class="h-100"
@@ -202,7 +202,7 @@
                 </oui-tile-definition>
             </oui-tile>
         </div>
-        <div class="col-xm-12 col-md-4 mb-5">
+        <div class="col-md-4 mb-5">
             <div class="mb-5" data-ng-if="$ctrl.screenshot">
                 <oui-spinner data-ng-if="$ctrl.loading.screenshot"></oui-spinner>
                 <a data-ng-href="http://{{$ctrl.domain.name}}"
@@ -283,7 +283,7 @@
         </div>
     </div>
     <div class="row d-lg-flex" data-ng-if="!$ctrl.domain.isExpired">
-        <div class="col-xm-12 col-md-4 mb-5">
+        <div class="col-md-4 mb-5">
             <div class="d-flex flex-column-reverse flex-lg-column">
                 <div data-wuc-guides
                      data-wuc-guides-list="'generalInformations'"
@@ -291,6 +291,30 @@
                      data-tr="tr">
                 </div>
             </div>
+        </div>
+        <div class="col-md-4 mb-5">
+            <oui-tile
+                class="h-100"
+                data-heading="{{ ::'domain_dashboard_options' | translate }}"
+                data-loading="$ctrl.loading.options">
+                <oui-tile-definition data-term="{{ ::'domain_dashboard_options_dns_anycast' | translate }}">
+                    <oui-tile-description>
+                        <span data-ng-if="$ctrl.options.dnsAnycast"
+                              data-translate="{{ 'domain_dashboard_options_status_' + ($ctrl.options.dnsAnycast.optionActivated ? 'enabled' : 'disabled') }}">
+                        </span>
+                        <span data-ng-if="!$ctrl.options.dnsAnycast"
+                              data-translate="domain_dashboard_options_unavailable">
+                        </span>
+                    </oui-tile-description>
+                    <oui-action-menu data-compact data-align="end"
+                                     data-ng-if="$ctrl.options.dnsAnycast && $ctrl.options.dnsAnycast.optionActivated">
+                        <oui-action-menu-item data-text="{{ ::'domain_dashboard_options_desactivate_option' | translate }}"
+                            aria-label="{{ ::'domain_dashboard_options_desactivate_option' | translate }}"
+                            data-on-click="setAction('options/delete/domain-options-delete', { domain: $ctrl.domain, option: $ctrl.options.dnsAnycast })">
+                        </oui-action-menu-item>
+                    </oui-action-menu>
+                </oui-tile-definition>
+            </oui-tile>
         </div>
     </div>
 </div>
--- a/client/app/domain/general-informations/domain-general-informations.controller.js
+++ b/client/app/domain/general-informations/domain-general-informations.controller.js
@@ -48,6 +48,7 @@ export default class DomainTabGeneralInformationsCtrl {
     this.hasStart10mOffer = false;
     this.isAllDom = this.$rootScope.currentSectionInformation === 'all_dom';
     this.isUK = _.last(this.domain.name.split('.')).toUpperCase() === 'UK';
+    this.options = {};
     this.loading = {
       allDom: false,
       associatedHosting: false,
@@ -57,6 +58,7 @@ export default class DomainTabGeneralInformationsCtrl {
       domainInfos: this.$scope.ctrlDomain.loading.domainInfos,
       changeOwner: false,
       whoIs: false,
+      options: false,
     };
     this.initActions();
     this.dnsStatus = {
@@ -102,6 +104,7 @@ export default class DomainTabGeneralInformationsCtrl {
     this.$scope.$on('domain.dnssec.lock.unlock.cancel', () => {
       this.vm.dnssec.uiSwitch.checked = !this.vm.dnssec.uiSwitch.checked;
     });
+    this.$scope.$on('Domain.Options.Delete', () => this.getAllOptionDetails(this.domain.name));
 
     if (!this.domain.isExpired) {
       this.getScreenshoot(this.domain.name);
@@ -114,6 +117,7 @@ export default class DomainTabGeneralInformationsCtrl {
     this.getAllNameServer(this.domain.name);
     this.getHostingInfos(this.domain.name);
     this.getAssociatedHosting(this.domain.name);
+    this.getAllOptionDetails(this.domain.name);
     this.updateOwnerUrl = this.getUpdateOwnerUrl(this.domain);
 
     this.getRules();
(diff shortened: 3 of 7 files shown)


client/app/domain/domain.service.js
```diff
@@ -1177,6 +1177,40 @@ angular.module('services').service(
...
     }
 
+    // --------------------- Options ----------------------------
+
+    /**
+     * Get all options
+     * @param {string} serviceName
+     */
+    getOptions(serviceName) {
+      return this.OvhHttp.get(`/domain/${serviceName}/option`, {
```
marie-j: Imo, those should be made through `ovh-api-services` to avoid using multiple data providers

client/app/domain/general-informations/domain-general-informations.controller.js
```diff
@@ -310,6 +314,33 @@ export default class DomainTabGeneralInformationsCtrl {
...
+          .then((optionDetail) => {
+            _.set(optionDetail, 'optionActivated', optionDetail.state === this.DOMAIN.DOMAIN_OPTION_STATUS.ACTIVE);
+            return optionDetail;
+          })),
+      ))
+      .then((options) => {
+        const transformedOptions = {};
+        _.forEach(options, (option) => {
+          transformedOptions[option.option] = option;
+        });
```
marie-j: could a `_.groupBy` suit the need here ? https://lodash.com/docs/4.17.11#groupBy

client/app/domain/options/delete/domain-options-delete.html
```diff
@@ -0,0 +1,15 @@
...
+         data-wizard-on-finish="deleteDomain"
+         data-wizard-title="('domain_tab_options_delete_'+ctrl.option.option+'_title') | translate">
+        <div data-wizard-step data-wizard-step-valid="!ctrl.loading">
+            <p data-ng-if="!ctrl.loading" data-translate="domain_tab_options_delete_confirm"></p>
+            <div class="text-center" data-ng-if="ctrl.loading">
+                <oui-spinner></oui-spinner>
+            </div>
+        </div>
+    </div>
+</div>
```
marie-j: Could this be turned into a `oui-modal` instead ?

client/app/domain/translations/Messages_fr_FR.json
```diff
@@ -817,5 +817,17 @@
...
\ No newline at end of file
+  "domain_dashboard_general_security": "Sécurité",
+  "domain_dashboard_options": "Résumé des options",
+  "domain_dashboard_options_dns_anycast": "DNS Anycast",
+  "domain_dashboard_options_status_enabled": "Activé",
+  "domain_dashboard_options_status_disabled": "Désactivé",
+  "domain_dashboard_options_unavailable": "Option non disponible",
+  "domain_dashboard_options_desactivate_option": "Résilier",
+  "domain_dashboard_options_fail": "Une erreur est survenue lors de la récupération des options de domaine",
+  "domain_tab_options_delete_confirm": "Voulez-vous vraiment supprimer le option ?",
```
marie-j:
```suggestion
  "domain_tab_options_delete_confirm": "Voulez-vous vraiment supprimer l'option ?",
```

client/app/domain/general-informations/domain-general-informations.controller.js
antleblanc: https://lodash.com/docs/3.10.1#groupBy ;)

client/app/domain/general-informations/GENERAL_INFORMATIONS.html
```diff
@@ -292,5 +292,30 @@
                 </div>
             </div>
         </div>
+        <div class="col-xm-12 col-md-4 mb-5">
```
jleveugle: `col-xm-12` is a mistake I think, and if you want to use `col-xs-12`, you should remove it, it's the default behaviour :-/

client/app/domain/options/delete/domain-options-delete.controller.js
```diff
@@ -0,0 +1,44 @@
...
+      return this.Domain.deleteOption(
+        this.domain.name,
+        this.option.option,
+      )
+        .then(() => this.Alerter.success(
+          this.$translate.instant('domain_tab_options_delete_success'),
+          this.$scope.alerts.main,
+        ))
+        .catch((err) => {
+          _.set(err, 'type', err.type || 'ERROR');
```
jleveugle: You should avoid updating `err`, which is an input (cf [PURE functions](https://www.nicoespeon.com/en/2015/01/pure-functions-javascript/))

For example, you can do something like that instead:

```javascript
this.Alerter.alertFromSWS(
	this.$translate.instant('domain_tab_options_delete_error'),
	Object.assign({}, err, { type: err.type || 'ERROR' }),
	this.$scope.alerts.main,
);
```

client/app/domain/options/delete/domain-options-delete.controller.js
```diff
@@ -0,0 +1,44 @@
...
+          this.Alerter.alertFromSWS(
+            this.$translate.instant('domain_tab_options_delete_error'),
+            err,
+            this.$scope.alerts.main,
+          );
+        })
+        .finally(() => {
+          this.loading = false;
+          this.$rootScope.$broadcast('Domain.Options.Delete');
```
jleveugle: You should avoid using `$rootScope`. `$scope.$emit` is sufficient in this case :)

Here, by doing $rootScope.$broadcast, you are going to broadcast the event to all scopes in the application.

By doing an `$emit`, you are going to dispatch the event only to the parent scopes :)

client/app/domain/general-informations/GENERAL_INFORMATIONS.html
```diff
@@ -292,5 +292,30 @@
...
+                        <span data-ng-if="!$ctrl.options.dnsAnycast"
+                              data-translate="domain_dashboard_options_unavailable">
+                        </span>
+                    </oui-tile-description>
+                    <oui-action-menu data-compact data-align="end"
+                                     data-ng-if="$ctrl.options.dnsAnycast && $ctrl.options.dnsAnycast.optionActivated">
+                        <oui-action-menu-item data-text="{{ ::'domain_dashboard_options_desactivate_option' | translate }}"
+                            aria-label="{{ ::'domain_dashboard_options_desactivate_option' | translate }}"
+                            data-on-click="setAction('options/delete/domain-options-delete', { domain: $ctrl.domain, option: $ctrl.options.dnsAnycast })"
+                            data-ng-if="$ctrl.options.dnsAnycast.optionActivated">
```
jleveugle: You test it on the parent element. Why retesting it here ?

client/app/domain/domain.service.js
ganeshkumar1989: All API calls in domain are done this way. So, I thought we can take this task of routing all the calls via ovh-api-services, as a separate story. Please let me know your thoughts :)

client/app/domain/general-informations/domain-general-informations.controller.js
ganeshkumar1989: I was actually looking for `_.keyBy` from https://lodash.com/docs/4.17.14#keyBy
Since this is not available in the version of lodash we are using, I thought I would do that manually :)

`_.groupBy` groups the elements of an array, where, each value in the returned object is an array. I just wanted each value in the returned object to be a single element from the array

client/app/domain/options/delete/domain-options-delete.html
ganeshkumar1989: Of course, yes. But all the modals in the domain section are now implemented this way. I think it is better to have a separate task to convert all of them to oui-modals at one shot. Please let me know your thoughts on this :)

client/app/domain/options/delete/domain-options-delete.controller.js
ganeshkumar1989: Understood @jleveugle . But in this case, we will need to catch the `Domain.Options.Delete` event in domain-general-informations.controller.js, which is not the parent scope. Hence, $emit is not going to work in this case. Please let me know if I understood it wrong :)

client/app/domain/general-informations/GENERAL_INFORMATIONS.html
ganeshkumar1989: You are right. This redundant check has been removed from the menu-item. Thanks :)

client/app/domain/translations/Messages_fr_FR.json
ganeshkumar1989: This is done :)

client/app/domain/general-informations/GENERAL_INFORMATIONS.html
ganeshkumar1989: This is remove. Thanks :)

client/app/domain/options/delete/domain-options-delete.controller.js
ganeshkumar1989: Thanks for that. This change is done

client/app/domain/domain.service.js
marie-j: This doesn't mean that we've got to continue that way. As we are adding new calls, it would be better starting uniformizing now, it's not much but it is also less to do later :)

client/app/domain/options/delete/domain-options-delete.html
marie-j: As said above, I think the sooner we improve our code the better, especially if it is new code

ganeshkumar1989: Agreed. This is done :)

client/app/domain/general-informations/domain-general-informations.controller.js
```diff
@@ -310,6 +314,33 @@ export default class DomainTabGeneralInformationsCtrl {
       });
   }
 
+  getAllOptionDetails(serviceName) {
+    this.loading.options = true;
+    return this.Domain.getOptions(serviceName)
+      .then(options => this.$q.all(
+        _.map(options, option => this.Domain.getOption(serviceName, option)
+          .then((optionDetail) => {
```
JeremyDec: You can do this to avoid parameter overriding :
```suggestion
          .then(optionDetail => Object.assign({}, optionDetail, { optionActivated: optionDetail.state === this.DOMAIN.DOMAIN_OPTION_STATUS.ACTIVE}
```

client/app/domain/options/delete/domain-options-delete.controller.js
```diff
@@ -0,0 +1,46 @@
+angular.module('controllers').controller(
+  'controllers.Domain.Options.Delete',
+  class DomainDnsLockCtrl {
```
JeremyDec: You could extract the class

client/app/domain/options/delete/domain-options-delete.controller.js
```diff
@@ -0,0 +1,46 @@
...
+    constructor($scope, $rootScope, $translate, Alerter, Domain) {
```
JeremyDec: Alphabetical order please

client/app/domain/options/delete/domain-options-delete.html
```diff
@@ -0,0 +1,16 @@
+<div data-ng-controller="controllers.Domain.Options.Delete as ctrl">
+    <oui-modal
+      data-heading="{{ ('domain_tab_options_delete_'+ctrl.option.option+'_title') | translate }}"
```
JeremyDec:
```suggestion
      data-heading="{{:: ('domain_tab_options_delete_'+ctrl.option.option+'_title') | translate }}"
```

client/app/domain/options/delete/domain-options-delete.html
```diff
@@ -0,0 +1,16 @@
...
+      data-primary-action="ctrl.deleteDomain()"
+      data-primary-disabled="ctrl.loading"
+      data-primary-label="{{ 'domain_tab_options_delete_confirm_button' | translate }}"
```
JeremyDec:
```suggestion
      data-primary-label="{{:: 'domain_tab_options_delete_confirm_button' | translate }}"
```

client/app/domain/options/delete/domain-options-delete.html
```diff
@@ -0,0 +1,16 @@
...
+      data-secondary-action="ctrl.closeModal()"
+      data-secondary-disabled="ctrl.loading"
+      data-secondary-label="{{ 'domain_tab_options_delete_cancel' | translate }}"
```
JeremyDec:
```suggestion
      data-secondary-label="{{:: 'domain_tab_options_delete_cancel' | translate }}"
```

client/app/domain/options/delete/domain-options-delete.html
```diff
@@ -0,0 +1,16 @@
...
+      data-on-dismiss="ctrl.closeModal()">
+        <p data-ng-if="!ctrl.loading" data-translate="domain_tab_options_delete_confirm"></p>
+        <div class="text-center" data-ng-if="ctrl.loading">
+            <oui-spinner></oui-spinner>
```
JeremyDec: You can remove this spinner, and use the `loading` property of `oui-modal`

jleveugle:
Hello @ganeshkumar1989,
Your pull request is now closed due to inactivity.
Feel free to re-open it if you have any update.
Anna Bot for OVH UX <3

client/app/domain/domain.service.js
ganeshkumar1989: This is done :)

client/app/domain/general-informations/GENERAL_INFORMATIONS.html
```diff
@@ -292,5 +292,29 @@
...
+                        <span data-ng-if="$ctrl.options.dnsAnycast"
+                              data-translate="{{ 'domain_dashboard_options_status_' + ($ctrl.options.dnsAnycast.optionActivated ? 'enabled' : 'disabled') }}">
+                        </span>
+                        <span data-ng-if="!$ctrl.options.dnsAnycast"
+                              data-translate="domain_dashboard_options_unavailable">
+                        </span>
+                    </oui-tile-description>
+                    <oui-action-menu data-compact data-align="end"
+                                     data-ng-if="$ctrl.options.dnsAnycast && $ctrl.options.dnsAnycast.optionActivated">
+                        <oui-action-menu-item data-text="{{ ::'domain_dashboard_options_desactivate_option' | translate }}"
```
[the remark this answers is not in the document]
ganeshkumar1989: This is done :)

client/app/domain/general-informations/domain-general-informations.controller.js
ganeshkumar1989: Thanks for that. This is done

client/app/domain/options/delete/domain-options-delete.controller.js
ganeshkumar1989: This class is not a reusable and is a controller class. I guess this class can be left as such?

client/app/domain/options/delete/domain-options-delete.controller.js
ganeshkumar1989: Done

client/app/domain/options/delete/domain-options-delete.html
ganeshkumar1989: Fixed :)

client/app/domain/options/delete/domain-options-delete.html
ganeshkumar1989: Fixed :)

client/app/domain/options/delete/domain-options-delete.html
ganeshkumar1989: Fixed :)

client/app/domain/options/delete/domain-options-delete.html
ganeshkumar1989: This is done

PR #1182, opened 2019-07-29
fix(anycast-dns): displayname punycode

varun257:
### Description of the Change

display domain name instead of id

### Applicable Issues

resolves MBP-416


client/app/domain/anycast/domain-dns-anycast.controller.js
```diff
@@ -12,7 +12,8 @@ angular.module('App').controller(
     }
 
     $onInit() {
-      this.domainName = this.$stateParams.productId;
+      this.domainName = this.$scope.ctrlDomain.domain.displayName;
```
Jisay: should it be safer to use `data` from a` setAction` call?

varun257: Passing the domain as parameter now.

PR #1211, opened 2019-08-23
fix(ie11): remove duplicate dependency and bump urijs

marie-j:
### Description of the Change

- use latest unqiue urijs version 

### Benefits

Prevent crash on IE11

MBE-252


package.json
```diff
@@ -148,7 +147,7 @@
     "raphael": "^2.2.8",
     "raven-js": "^3.27.1",
     "request": "^2.51.0",
-    "uri.js": "^0.1.3",
+    "urijs": "^1.19.1",
```
FredericEspiau: No breaking changes ? In case they don't respect semantic versioning

antleblanc: Nothing that impact the application codebase.

PR #1223, opened 2019-08-28
fix(domain.dns): allow to order dns anycast

marie-j:
DTRUN-2893, MANAGER-3217


antleblanc:
@marie-j Shouldn't target `master` branch?

PR #1224, opened 2019-08-28
fix: remove ng-ovh-otrs dependency

antleblanc:
# Remove ng-ovh-otrs dependency

Since the OtrsPopup can only be accessible from the Dedicated Control panel,
there is no reason to keep the `otrs-popup` directive.

## :bug: Bug Fix

8708fdd - fix: remove ng-ovh-otrs dependency

## :link: Related

MANAGER-2710

## :house: Internal

- No QC required.

PR #1234, opened 2019-09-10
fix: restore OTRS usage

FredericEspiau:


antleblanc:
Looks like a revert from #1224 (apologies about it if it causes some troubles).

But it would be great to get a more in depth about why this dependency is still needed for the Web Control Panel.

Thanks!

